PR #1, opened 2015-05-14
Update some links in README.md

martijnvermaat:
Not sure if the GitHub one is now supposed to become the primary repository for biopet, but this update is based on that assumption.

Welcome by the way :smiley: :green_heart:


wyleung:
:) this is not our primary repo, more for the public to report issues and such.. thanks for the PR. We need to figure our what the best setup would be for the markdown files.

PR #2, opened 2017-01-14
Fix BIOPET-498

ffinfo:


sndrtj (changes requested):
Looks good :+1: One small tidbit though: IIRC non-empty directories will not get deleted by `deleteOnExit`. You may want to ensure the files inside the temporary output directory get removed on teardown.

ffinfo:
Good point, changed this

sndrtj (approved):

PR #4, opened 2017-01-14
Fix BIOPET-379

ffinfo:


biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/MultiSampleQScript.scala
```diff
@@ -141,11 +141,17 @@ trait MultiSampleQScript extends SummaryQScript { qscript: QScript =>
...
     /** returns a set with library names or throws error when not found */
     protected def libIds: Set[String] = {
-      val ids = ConfigUtils.getMapFromPath(globalConfig.map, List("samples", sampleId, "libraries")).getOrElse(Map()).keySet
-      if (ids.isEmpty) {
-        throw new IllegalStateException(s"No libraries found in config for sample $sampleId")
+      val ids: Set[String] = try {
+        ConfigUtils.getMapFromPath(globalConfig.map, List("samples", sampleId, "libraries")).getOrElse(Map()).keySet
+      } catch {
+        case e: IllegalStateException if e.getMessage == "Value is not a map: library" =>
+          Logging.addError("libraries is not a map")
```
sndrtj: This is a somewhat nebulous error message if one is not familiar with the internal structure we use here. Perhaps simply state the contents are wrongly formatted?

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
```diff
@@ -177,4 +177,52 @@ object BiopetQScript {
...
+
+  def safeFailFiles(function: QFunction): Seq[File] = {
+    try {
+      function.failOutputs
+    } catch {
+      case e: NullPointerException => Seq()
+    }
+  }
+
+  def safeIsDone(function: QFunction): Boolean = {
```
sndrtj: I would change this function and `safeIsFail` to have type `Option[Boolean]`. You return simply `false` here when the underlying method does not exist, but I believe that's potentially giving a wrong answer. Returning `None` (as in, no boolean exists here) is semantically better, imho.

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/MultiSampleQScriptTest.scala
```diff
@@ -272,7 +276,7 @@ object MultiSampleQScriptTest {
       def summaryFiles: Map[String, File] = Map()
 
       /** Name of summary output file */
-      def summaryFile: File = null
+      def summaryFile: File = new File("./summary.json")
```
sndrtj, line 279: If this file gets created during the test, we want to remove it after teardown.

sndrtj (changes requested):
I have several comments you may want to look at on this PR.

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/MultiSampleQScriptTest.scala
ffinfo: is does not get created :)

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
ffinfo: maybe indeed a better option

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/MultiSampleQScript.scala
ffinfo: changed this

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
ffinfo: Changed this

sndrtj (approved):

PR #6, opened 2017-01-18
Fix for error when creating the report

papanikos:


ffinfo (approved):

sndrtj (commented):
Why should we no longer require this variable to have a value?

papanikos:
Some samples might have 0 reads mapped and that causes the pipeline to error. This fix handles these cases.

PR #9, opened 2017-01-22
Udate Bios branch to latest develop (with SLurm support)

ffinfo:


ffinfo:
@sndrtj: no need for review here, this all about code that is already merged in Develop anyway.

PR #13, opened 2017-01-25
Adding unassigned reads to krona plot

ffinfo:


ffinfo:
The failing test here is fixed in https://github.com/biopet/biopet/pull/12

sndrtj:
Re-running tests after merging #12.

sndrtj (approved):

PR #14, opened 2017-01-27
added a short text to user documentation

akaljuvee:
Added a short text to user documentation, it says that it's possible to generate config files with the help of Biopet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ node('local') {
         }
 
         stage('Check Documentation') {
-            sh 'mkdocs build'
+            sh 'mkdocs build --clean --strict'
         }
 
         if(currentBuild.result == null || "SUCCESS".equals(currentBuild.result)) {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -17,10 +17,14 @@ node('local') {
             }
         }
 
-        stage('Report tests') {
+        stage('Report Tests') {
             junit '*/target/surefire-reports/*.xml'
         }
 
+        stage('Check Documentation') {
+            sh 'mkdocs build'
```
ffinfo: Can this be done with --clean and --strict ? Without this mkdocs will not give a non-0 exitcode and Jenkins will still accept this.

ffinfo (changes requested):

Jenkinsfile
akaljuvee: indeed the flag --strict should be present, but this mkdocs isn't consistent - if there's a wrong *.md file written in mkdocs.yml then it's an error, but if there's a mistake in one of the *.md files then it's a warning

ffinfo (approved):

PR #17, opened 2017-02-02
one log line

akaljuvee:


ffinfo (changes requested):
Is it also possible to print the total number of jobs?

ffinfo (approved):

PR #20, opened 2017-02-07
Disable merging for specific variantcallers

ffinfo:


biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
```diff
@@ -107,6 +107,7 @@ trait BiopetQScript extends Configurable with GatkLogging { qscript: QScript =>
       count += 1
       if (count % 500 == 0) logger.info(s"Preprocessing done for ${count} jobs out of ${functions.length} total")
     }
+    logger.info(s"Preprocessing done for ${functions.length} funtions")
```
sndrtj: There's a tiny typo here: `funtions` -> `functions`.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
```diff
@@ -121,6 +122,7 @@ trait BiopetQScript extends Configurable with GatkLogging { qscript: QScript =>
       if (!i.file.isAbsolute) Logging.addError(s"Input file should be an absolute path: ${i.file}")
     }
 
+    logger.info("Set stdout file when not set")
```
sndrtj, line 125: I assume you want to say that `stdout` is automatically set because it is not explicitly set by the user?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
```diff
@@ -112,21 +112,22 @@ class ShivaVariantcalling(val root: Configurable) extends QScript
...
         vtDecompose.outputVcf = swapExt(caller.outputDir, caller.outputFile, ".vcf.gz", ".decompose.vcf.gz")
         add(vtDecompose, Tabix(this, vtDecompose.outputVcf))
-        cv.variant :+= TaggedFile(vtDecompose.outputVcf, caller.name)
-      } else cv.variant :+= TaggedFile(caller.outputFile, caller.name)
+        if (caller.mergeVcfResults) cv.variant :+= TaggedFile(vtDecompose.outputVcf, caller.name)
+      } else if (caller.mergeVcfResults) cv.variant :+= TaggedFile(caller.outputFile, caller.name)
+    }
+    if (cv.variant.nonEmpty) {
+      add(cv)
+      addStats(finalFile, "final")
```
sndrtj, line 130: I think we should also collect stats for the cases when callsets do not get merged. In that case, there are two possibilities:

* There is only one callset, because only one caller was used, and the final file is simply the only VCF file. 
* There are multiple callsets, and stats should preferably be collected for all of them.

sndrtj (changes requested):
How does this play with Shiva runs with Toucan annotation enabled? IIRC, Toucan can only handle one single input VCF file. 

That means that we can either make Toucan accept multiple VCF files, or call Toucan in Shiva several times. The latter is likely easier to implement. That said, I think Toucan+Varda+merge_vcf_results=false should be impossible and caught before attempting to run. Putting multiple callsets of the same individual in a Varda database is going to send the frequencies off.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
ffinfo: it mean when not set by the pipeline ;). Maybe change this to debug?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
ffinfo: This is already done, each variantcaller executes it's own vcfstats and in the end there is a vcfstats on the final file.

ffinfo:
just an example how to use this:
``` yml
variantcallers:
- haplotypecaller
- haplotypecaller_gvcf
- raw
vcfraw:
  merge_vcf_results: false
```

In this case 3 variantcallers are executed but the final vcf file that will go to toucan only contain haplotypecaller and haplotypecaller_gvcf. Still for al 3 variantcallers the separated vcf is still there.

Saying this maybe I should add a requirement that at least 1 variantcaller should end up in the final file. Otherwise the final vcf will not even by created while toucan does expect it.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
ffinfo: Fixed

sndrtj:
Could you add that example (or something similar) to the docs as well? Would be very useful. Thanks :-).

sndrtj (approved):

PR #24, opened 2017-02-27
Fix biopet 546

ffinfo:
This PR is the first for the summary branch.

Unit testing does not work completely but will be fixed in an other merge to this branch


biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/implicits/Either.scala
```diff
@@ -0,0 +1,40 @@
+package nl.lumc.sasc.biopet.utils.implicits
+
+/**
+ * Created by pjvanthof on 20/02/2017.
+ */
+object Either {
```
bow, line 6: If I recall correctly, many of the implicits here are already defined in the scalaz library (using a slightly different type, but scalaz is already a dependency of ours anyway). Plus, the mapping convention is to use `Left` for error messages and `Right` for the results. Are there cases where we map `Option` to `Left` in our codebase?

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
```diff
@@ -0,0 +1,146 @@
...
+import java.sql.Date
+
+import slick.driver.H2Driver.api._
+
+/**
+ * Created by pjvan_thof on 27-1-17.
+ */
+object Schema {
+
+  case class Run(id: Int, name: String, outputDir: String, version: String, commitHash: String,
```
bow, line 12: More documentations would be nice, for example:

* Are there any forbidden characters in `name`? How about empty strings? If the validation is not done here, where is it done?
* What is an example of a `version` string? Do we accept anything or do we expect a pattern?
* Are there any limitations on the commit hash? Is any number of characters fine?

This is for all the classes (tables) defined in this `Schema` object.

bow (changes requested):

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
ffinfo: Empty string I never tested but I assume this just works. If the database implementation does not accept this it will raise an error but in theory a empty string is still a string.

The schema here is very basic because a lot of those things are not working in sqlite. If using this for in an other implementation we can extends the schema's to a more specific implementation.

For a version there is no regex because not all tools give the same kind of version numbers.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/implicits/Either.scala
ffinfo: I can take a look into this indeed

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
bow: Yes, this is all fine ~ just needs to be noted in some scaladoc in the code :).

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/implicits/Either.scala
bow: Great. For the record, scala'z `Either` is called `\/`. It's not exactly pronounceable, but the official term is disjunction (like the mathematical disjunction symbol ν, which is sort of a fancy name for 'or'.

Anyway, look for `Option.to[Right|Left]Disjunction` methods there.

bow:
By the way, also a general comment (since the code is not fully merged here yet but I already see traces of it):

I see a lot of `Duration.Inf` as argument in the `Await` calls, whose use I think should be minimized (otherwise it will be hard to debug when there is a hanging error somewhere, for example).

In general I would keep the `Future[_]` return types as long as possible (pass it on to the caller up until the value must be extracted) and not call `Await` that much. For example, is it possible to call `Await` only at the moment when the HTML file is about to be written? That way you only have a single `Await` call whose duration you can probably configure via a command line argument.

If you need to sequence multiple `Future` calls, there are also ways to do this. You can almost treat it like an `Option` (almost ~ a better type is actually called `Task` in Scalaz), so long as you're not doing weird things inside the future and keep side effects to a minimum.

Anyway, I'll show more concrete examples when the other parts of the code are merged into this branch.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
sndrtj: > but I assume this just works.

I would be careful with this. Adding a unit test for this should be possible, right?

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
```diff
@@ -0,0 +1,146 @@
...
+  val modules = TableQuery[Modules]
+
+  case class Stat(runId: Int, pipelineId: Int, moduleId: Option[Int], sampleId: Option[Int], library: Option[Int], content: String)
+  class Stats(tag: Tag) extends Table[Stat](tag, "Stats") {
+    def runId = column[Int]("runId")
+    def pipelineId = column[Int]("pipelineId")
+    def moduleId = column[Option[Int]]("moduleId")
+    def sampleId = column[Option[Int]]("sampleId")
+    def libraryId = column[Option[Int]]("libraryId")
+    def content = column[String]("content")
```
sndrtj, line 85: Shouldn't the content technically be nullable, as this is highly variable?

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
```diff
@@ -0,0 +1,146 @@
...
+    def runId = column[Int]("runId")
+    def pipelineId = column[Int]("pipelineId")
+    def moduleId = column[Option[Int]]("moduleId")
+    def sampleId = column[Option[Int]]("sampleId")
+    def libraryId = column[Option[Int]]("libraryId")
+    def key = column[String]("key")
+    def path = column[String]("path") // This should be relative to the outputDir
+    def md5 = column[String]("md5")
+    def link = column[Boolean]("link", O.Default(false))
+    def size = column[Long]("size")
```
sndrtj, line 119: I'm not quite sure whether this is a useful metric to contain in the summary.

sndrtj (commented):
I still have some questions regarding the refactor of `Configurable.root` to `Configurable.parent`. This is a breaking API change that requires a modification on all implementing objects. As the new `Configurable.root` is only accessed 5 times, I am not entirely sure of the benefit either. 

My suggestion would be to name the new `Configurable.root` something else (`ultimateRoot` or so?), and change `parent` back to `root`. 

If we do keep the refactor, we should at least document it in the developer documentation.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
ffinfo: if it's empty it becomes "{}"

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/implicits/Either.scala
ffinfo: Did make a issue in this epic for to look into this. For this merge is can be ignored, will be in the next batch.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
ffinfo: @sndrtj: yes indeed, will also be added in the next batch

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/Schema.scala
ffinfo: well this give you a quick overview of the size of the project within the database. it's easy to add and much faster then `du`

PR #31, opened 2017-03-02
Fix biopet 587

ffinfo:


bammetrics/src/main/scala/nl/lumc/sasc/biopet/pipelines/bammetrics/BammetricsReport.scala
```diff
@@ -175,8 +177,8 @@ object BammetricsReport extends ReportBuilder {
                            statsPaths: Map[String, List[String]],
                            xKey: String,
                            yKey: String,
-                           pipeline: Either[Int, String],
-                           module: Option[Either[Int, String]],
+                           pipeline: \/[Int, String],
+                           module: Option[\/[Int, String]],
```
bow: This function (`writePlotFromSummary`) needs more documentation. It's not immediately clear what the arguments are, especially since we have `Int \/ String` here. Does the `Int` there denote error code, for example?

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/report/MultisampleReportBuilderTest.scala
```diff
@@ -41,10 +45,25 @@ class MultisampleReportBuilderTest extends TestNGSuite with Matchers {
...
+    val dbFile = File.createTempFile("summary.", ".db")
+    dbFile.deleteOnExit()
+    val db = SummaryDb.openSqliteSummary(dbFile)
+    db.createTables()
+
+    val sample = Some("sampleName")
+    val lib = Some("libName")
+
+    sample.foreach { sampleName =>
+      val sampleId = Await.result(db.createSample(sampleName, 0), Duration.Inf)
```
bow, line 57: This is a test, I realize, but `createSample` is being called in a `.foreach` loop here and immediately used for `Await`. This means the `Futures` are actually launched one after another, so you don't get the benefit of async here.

bow (changes requested):
Hey @ffinfo,

I did a quick check on the changes, but I think there are several things that need to be discussed more, mainly on the `Future` usage and also using scalaz's `left` and `right`. I prefer to do this with @sndrtj and anyone else who wants to join, sometime next week (monday 1 pm)?

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/report/MultisampleReportBuilderTest.scala
ffinfo: Not completely true. By doing Await the current thread is mark open for other tasks. So if your program is running a lot of tasks/futures/.par this is being done meanwhile.

For the testing this is indeed not going in parallel but as you already mention for testing this is less important.

Where is being used the most in the report. Here each page is already there own thread/task. So doing Await is only positive so the cpu is free to do other stuff ;).

bammetrics/src/main/scala/nl/lumc/sasc/biopet/pipelines/bammetrics/BammetricsReport.scala
ffinfo: Agree, will add more documentation

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/report/MultisampleReportBuilderTest.scala
bow: Hmm..you can try this on your own intepreter: compare the execution time of the two code blocks below.

(but first, some boilerplate)
```scala
import scala.concurrent._
import scala.concurrent.duration._
implicit def ec = ExceutionContext.global._
```

Now for the `Future` function that has a noticeable delay:
```scala
def func1(n: int) = Future { Thread.sleep(3000); n * 2 }
```

Execution with the awat inside `.foreach` is something like this:
```scala
Seq(1, 2, 300).foreach { num =>
   val result = Await.result(func1(num), 20.seconds)
   println(result)
}
```

And with `Await` launched just once:
```scala
val futures = Seq(1, 2, 300).map { func1 }
val future = Future.sequence(futures)
val result = Await.result(future, 20.seconds)
result.foreach { println }
```

The first block should take roughly about 9 seconds (since `func1` takes 3 seconds to complete and we have 3 invocations). But the second one should only take a little more than 3 seconds, since we only do the `Await` call once.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/summary/SummaryQScript.scala
```diff
@@ -58,40 +58,23 @@ trait SummaryQScript extends BiopetQScript { qscript: QScript =>
...
-  }
-
-  /**
-   * Add a module to summary for this pipeline
-   *
    * @param summarizable summarizable to add to summary for this pipeline
    * @param name Name of module
    * @param sampleId Id of sample
    * @param libraryId Id of libary
+   * @param forceSingle If true it replaces summarizable instead of adding to it
```
sndrtj, line 65: Calling this `replace` may be a bit clearer as to to what this does.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/summary/WriteSummary.scala
```diff
@@ -159,15 +166,21 @@ class WriteSummary(val parent: SummaryQScript) extends InProcessFunction with Co
...
-      case f: BiopetCommandLineFunction with Version =>
-        Some(db.createOrUpdateExecutable(qscript.summaryRunId, f.configNamespace, f.getVersion, Option(f.executable)))
-      case f: Configurable with Version =>
-        Some(db.createOrUpdateExecutable(qscript.summaryRunId, f.configNamespace, f.getVersion))
-      case _ => None
+    val pipeFunctions = (for (f <- qscript.functions) yield f match {
+      case f: BiopetCommandLineFunction => f.pipesJobs
+      case _                            => Nil
+    }).flatten
+    (for (f <- qscript.functions ++ pipeFunctions) yield {
```
sndrtj, line 173: You are already looping through all `qscript.functions` above. Perhaps there's a a way of doing it all in one loop.

biopet-core/src/test/scala/nl/lumc/sasc/biopet/core/summary/WriteSummaryTest.scala
```diff
@@ -41,22 +46,19 @@ class WriteSummaryTest extends TestNGSuite with Matchers {
...
-    summary.getValue(sampleId, libId, name, "executables").get shouldBe a[Map[_, _]]
-
-    summary.getValue("meta") should not be None
-    summary.getValue("meta", "pipeline_name") shouldBe Some(name)
-    summary.getValue("meta", "last_commit_hash") shouldBe Some(nl.lumc.sasc.biopet.LastCommitHash)
-    summary.getValue("meta", "pipeline_version") shouldBe Some(nl.lumc.sasc.biopet.Version)
-    summary.getValue("meta", "output_dir") shouldBe Some(new File(".").getAbsolutePath)
-    summary.getValue("meta", "summary_creation") should not be None
+    val run = Await.result(summary.getRuns(runId = Some(runId)), Duration.Inf).head
+    run.commitHash shouldBe nl.lumc.sasc.biopet.LastCommitHash
```
sndrtj, line 58: :+1:

flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/FlexiprepReport.scala
```diff
@@ -43,9 +44,9 @@ object FlexiprepReport extends ReportBuilder {
     val flexiprepPage = this.flexiprepPage
     ReportPage(List("Versions" -> ReportPage(List(), List("Executables" -> ReportSection("/nl/lumc/sasc/biopet/core/report/executables.ssp"
     )), Map()),
-      "Files" -> ReportPage(List(), List(
-        "Input fastq files" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/flexiprep/flexiprepInputfiles.ssp"),
-        "After QC fastq files" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/flexiprep/flexiprepOutputfiles.ssp")
+      "Files" -> ReportPage(List(), List( //TODO: Fix files
```
sndrtj, line 47: Can this be in this merge request?

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/summary/WriteSummary.scala
ffinfo: Tried that but makes it a lot more complex

flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/FlexiprepReport.scala
ffinfo: yes, still an issue for this open. Will be fixed before merge to develop

PR #35, opened 2017-03-09
Test fail on format/git issues

ffinfo:


akaljuvee:
I thought that building with mvn creates lots of files that aren't in the repository. Isn't it so? Git diff won't count those? What is the one file that gets created that this diff is expected to catch?

ffinfo:
The target dirs are in the .gitignore file. So those files are not included in the diff.

akaljuvee:
ok, but what is this one bad file that gets created? I'd also say it in the error msg why the checkout isn't clean anymore.

akaljuvee (approved):

PR #37, opened 2017-03-20
Adding more output files to bamstats

ffinfo:
Fixing BIOPET-621 with this


rhpvorderman (changes requested):
The code is clear and understandable. But for some reason Jenkins gave this error:
Cannot run program "Rscript": error=2, No such file or directory

Cannot see how that stems from using this code?

rhpvorderman (approved):

PR #38, opened 2017-03-20
Fix biopet 607

rhpvorderman:
Autogenerated some missing documentation.


biopet-autodoc/synchronize_docs.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+
```
sndrtj, line 2: For some extra checking, you may want to set `set -eu` and `set -o pipefail` in the beginning of the script. This makes sure bash ends when it encounters an error or an unset variable.

biopet-autodoc/templateparser.py
```diff
@@ -0,0 +1,40 @@
...
+from jinja2 import Environment, FileSystemLoader
+import os
+import argparse
+import ast
+import yaml
+
+if __name__ == "__main__":
+    
+    #Block with command line options
+    parser=argparse.ArgumentParser(description="")
```
sndrtj, line 13: The `description` is an optional variable. If you set it to an empty string, you may as well leave it out altogether ;-).

biopet-autodoc/templateparser.py
```diff
@@ -0,0 +1,40 @@
...
+    parser.add_argument("-t", type=str, nargs=1, help="Templatefile to be used")
```
sndrtj, line 14: If you don't set `nargs`, accessing this value becomes easier, as you don't have to access a list.

biopet-autodoc/templateparser.py
```diff
@@ -0,0 +1,40 @@
...
+    tool['run']=arguments.tool_run
+    help_file=open(arguments.tool_option_list, 'r')
+    tool['option_list']=help_file.read()   #This reads the option list file to a raw string
+
+    # Open jinja2 template and render it with variables
+    j2_env = Environment(loader=FileSystemLoader("/"))
+    rendered = j2_env.get_template(template).render(tool=tool)
+
+    # Write the template results to the output file
+    file=open(output_file, "w")
```
sndrtj, line 37: The `seek` call here is unnecessary. Also, if you put this under a context handler, you do not have to manually close the handle.
E.g.:

```
with open(output_file, "w") as output_handle:
    output_handle.write(rendered)
```

sndrtj (changes requested):
Absolutely great idea! :+1: 

I do have some small comments on the scripts. Please check them out :-).

rhpvorderman:
The autodocumentation will be done in another way.

PR #42, opened 2017-03-23
Fix biopet 587

ffinfo:


bammetrics/src/main/resources/nl/lumc/sasc/biopet/pipelines/bammetrics/covstatsPlot.ssp
```diff
@@ -1,50 +1,46 @@
...
 #import(java.io.File)
-<%@ var summary: Summary %>
-<%@ var sampleId: Option[String] %>
-<%@ var libId: Option[String] = None %>
+<%@ var summary: SummaryDb %>
+<%@ var sampleId: Option[Int] %>
+<%@ var libId: Option[Int] = None %>
 <%@ var outputDir: File %>
 <%@ var metricsTag: String = "bammetrics" %>
 <%@ var target: Option[String] %>
```
sndrtj, line 16: It appears `target` is assumed to be `Some(String)` everywhere in this file. If that's the case, it'd be better to simply change the type to simply `String`.

bammetrics/src/main/resources/nl/lumc/sasc/biopet/pipelines/bammetrics/rnaHistogram.ssp
```diff
@@ -62,27 +61,25 @@
         #end
     </tr></thead>
     <tbody>
-        #for (sample <- samples.toList.sorted)
+        #for (sample <- samples.sortBy(_.name))
```
sndrtj, line 64: This section gets repeated in several places. To prevent code duplications, perhaps it's better to place a function for this in utils.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/SummaryDb.scala
```diff
@@ -448,6 +346,138 @@ class SummaryDb(val db: Database) extends Closeable {
     db.run(filesFilter(Some(runId), Some(pipeline), Some(module), Some(sample), Some(library), Some(key)).result).map(_.headOption)
   }
 
+  /** Returns a [[Query]] for [[Executables]] */
+  def executablesFilter(runId: Option[Int], toolName: Option[String]) = {
```
sndrtj: Missing return type.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/SummaryDb.scala
```diff
@@ -448,6 +346,138 @@ class SummaryDb(val db: Database) extends Closeable {
...
+    getLibraryId(runId, sampleId, name).flatMap(_ match {
+      case Some(id: Int) =>
+        db.run(libraries.filter(_.name === name).filter(_.id === id).filter(_.sampleId === sampleId).map(_.tags).update(tags))
+          .map(_ => id)
+      case _ => createLibrary(name, runId, sampleId, tags)
+    })
+  }
+
+  /** Creates a new pipeline, even if it already exist. This may give a database exeption */
+  def forceCreatePipeline(name: String, runId: Int): Future[Int] = {
```
sndrtj: Exactly which exceptions may this throw?

carp/src/main/scala/nl/lumc/sasc/biopet/pipelines/carp/Carp.scala
```diff
@@ -82,8 +82,6 @@ class Carp(val parent: Configurable) extends QScript with MultisampleMappingTrai
       bamMetricsFilter.summaryName = "bammetrics-filter"
       addSummaryQScript(bamMetricsFilter)
```
sndrtj, line 84: Why is this removed?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
```diff
@@ -105,13 +106,14 @@ trait ShivaReportTrait extends MultisampleMappingReportTrait {
...
 
   /** Files page, can be used general or at sample level */
-  override def filesPage(sampleId: Option[Int] = None, libraryId: Option[Int] = None): ReportPage = {
-    val vcfFilesSection = if (variantcallingExecuted) List("VCF files" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/shiva/outputVcfFiles.ssp",
-      Map("sampleId" -> sampleId)))
-    else Nil
-    val oldPage = super.filesPage(sampleId, libraryId)
-    oldPage.copy(sections = oldPage.sections ++ vcfFilesSection)
-  }
+  //TODO: fix files pages
```
sndrtj: Still some commented-out code.

biopet-core/src/main/resources/nl/lumc/sasc/biopet/utils/rscript/plotScatter.R
```diff
@@ -1,40 +0,0 @@
-library(reshape2)
```
sndrtj, line 1: Is the functionality of these R scripts taken over by something else? If not, that leaves us with fewer plots.

sndrtj (changes requested):

biopet-core/src/main/resources/nl/lumc/sasc/biopet/utils/rscript/plotScatter.R
ffinfo: yes, those 3 script where duplicated in biopet-core while they already exist in biopet-utils.

bammetrics/src/main/resources/nl/lumc/sasc/biopet/pipelines/bammetrics/covstatsPlot.ssp
ffinfo: Yes i do see your point. This is done because this same var is used in other templates where target is optional. Because the var are defined at a certain branch point of the page structure this type/var should be the same is all underlying pages.
For example for target there is also a variantSummary plot that is the same template as the general plot without a target.

bammetrics/src/main/resources/nl/lumc/sasc/biopet/pipelines/bammetrics/rnaHistogram.ssp
ffinfo: I agree there is some code duplication. To fix this all templates should be rearranged. This is a bit related to BIOPET-628. Because this change is to large to do in the current issue I will make a new issue for this to look into.

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/SummaryDb.scala
ffinfo: This deepens on the database implementation that is used. So this is not in our control, I wise it was.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
ffinfo: ah yes ty, this part was generalized but forgot to remove this from Shiva. I did remove this

carp/src/main/scala/nl/lumc/sasc/biopet/pipelines/carp/Carp.scala
ffinfo: This was duplicated, this is already executed in MultiSampleMapping

mapping/src/main/scala/nl/lumc/sasc/biopet/pipelines/mapping/MultisampleMappingReport.scala
```diff
@@ -114,10 +114,29 @@ trait MultisampleMappingReportTrait extends MultisampleReportBuilder {
...
+    val dbFiles = summary.getFiles(runId, sample = sampleId.map(SampleId),
+      library = libraryId.map(LibraryId))
+      .map(_.groupBy(_.pipelineId))
+    val modulePages = dbFiles.map(_.map {
+      case (pipelineId, files) =>
+        val moduleSections = files.groupBy(_.moduleId).map {
+          case (moduleId, files) =>
+            val moduleName: Future[String] = moduleId match {
+              case Some(id) => summary.getModuleName(pipelineId, id).map(_.getOrElse("Pipeline"))
+              case _        => Future("Pipeline")
```
bow, line 127: Is this meant to be a completed `Future`? If so, `Future.successful` is more straightforward since it does not create any extra `Future` that evaluates its argument.

mapping/src/main/scala/nl/lumc/sasc/biopet/pipelines/mapping/MultisampleMappingReport.scala
```diff
@@ -114,10 +114,29 @@ trait MultisampleMappingReportTrait extends MultisampleReportBuilder {
   }
 
   /** Files page, can be used general or at sample level */
-  def filesPage(sampleId: Option[Int] = None, libraryId: Option[Int] = None): ReportPage = {
-    val flexiprepExecuted = Await.result(summary.getStatsSize(runId, "flexiprep", NoModule, mustHaveLibrary = true), Duration.Inf) >= 1
-
-    ReportPage(List(), Nil, Map())
+  def filesPage(sampleId: Option[Int] = None, libraryId: Option[Int] = None): Future[ReportPage] = {
```
bow, line 117: In this one function I see several await calls even though the return type is already `Future`. Is it not possible to use `Future.flatMap` or `Future.map` instead? So you can do `Await` on whoever calls `filesPage` instead?

bow (changes requested):

mapping/src/main/scala/nl/lumc/sasc/biopet/pipelines/mapping/MultisampleMappingReport.scala
ffinfo: This is know and get solved in BIOPET-618

mapping/src/main/scala/nl/lumc/sasc/biopet/pipelines/mapping/MultisampleMappingReport.scala
ffinfo: This is know and get solved in BIOPET-618

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/summary/db/SummaryDb.scala
ffinfo: Just did a last commit to fix this

PR #49, opened 2017-03-29
Fix pipes

ffinfo:
This changes comes from the bios branch but should also happen in develop


biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/samtools/SamtoolsView.scala
```diff
@@ -34,17 +34,19 @@ class SamtoolsView(val root: Configurable) extends Samtools {
   var F: List[String] = config("F", default = List.empty[String])
 
   def cmdBase = required(executable) +
+  @Input(required = false)
+  var L: Option[File] = None
+
```
johnmous: /** Returns command to execute */

johnmous (approved):

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/samtools/SamtoolsView.scala
ffinfo: ahh, yes ty, added that back

PR #52, opened 2017-03-30
Single-sample jobs for count file generation

sndrtj:


tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +37,76 @@ class Tarmac(val root: Configurable) extends QScript with PedigreeQscript with S
...
+     */
+    lazy val outputXhmmCountFile: String \/ File = {
+      val outFile = new File(sampleDir + File.separator + s"$name.dcov")
+      (inputXhmmCountFile, bamFile) match {
+        case (Some(f), _) => {
+          val ln = new Ln(root)
+          ln.input = f
+          ln.output = outFile
+          add(ln)
+          \/-(ln.output)
```
bow: This is more of a comment about Scalaz's `\/` case class use. You can also use `ln.output.right` to make it more legible. The same applies for the `.left` use as well. Sometimes you need to add the type of the other side (e.g. `ln.output.right[String]` in this case), but only when there is inadequate type information (which isn't the case here).

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +37,76 @@ class Tarmac(val root: Configurable) extends QScript with PedigreeQscript with S
...
+        }
+        case _ => -\/(s"Cannot find bam file or wisecondor count for sample" +
+          s" $name. At least one must be given.")
+      }
+
+    }
+
     /** Function to add sample jobs */
-    def addJobs(): Unit = {}
+    def addJobs(): Unit = {
```
bow, line 123: You can also just make `outputWisecondorCountFile` just create the job object instead of adding it. It is `addJobs` who would then add the output, so you have one single place where all the side-effecting function (the job addition function) is done.

ffinfo: agree on that, add should only be done inside addJobs functions

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +37,76 @@ class Tarmac(val root: Configurable) extends QScript with PedigreeQscript with S
...
+     * Create XHMM count file or create link to input count file
+     * Precedence is given to existing count files.
+     * Returns a disjunction where right is the file, and left is
+     * a potential error message
+     */
+    lazy val outputXhmmCountFile: String \/ File = {
+      val outFile = new File(sampleDir + File.separator + s"$name.dcov")
+      (inputXhmmCountFile, bamFile) match {
+        case (Some(f), _) => {
+          val ln = new Ln(root)
```
ffinfo: Why using Ln here? maybe better to just use the file directly? (or is this not allowed by the tool? ;) )

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +37,76 @@ class Tarmac(val root: Configurable) extends QScript with PedigreeQscript with S
...
+    val inputWisecondorCountFile: Option[File] = config("wisecondor_count_file")
     val bamFile: Option[File] = config("bam")
 
+    /**
+     * Create XHMM count file or create link to input count file
+     * Precedence is given to existing count files.
+     * Returns a disjunction where right is the file, and left is
+     * a potential error message
+     */
+    lazy val outputXhmmCountFile: String \/ File = {
```
ffinfo: Don't think using \/ here is really usefull, just use File and use Logging.addError inside here

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +37,76 @@ class Tarmac(val root: Configurable) extends QScript with PedigreeQscript with S
...
+      val outFile = new File(sampleDir + File.separator + s"$name.dcov")
```
ffinfo: this is maybe more efficient: `new File(sampleDir, s"$name.dcov")`

ffinfo (changes requested):
Maybe a general comment, you should update this branch to develop, summary changes are in there now

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
sndrtj: This was addressed in b9bdbc9

sndrtj:
https://github.com/biopet/biopet/pull/53 merges develop into tarmac.

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -29,13 +38,90 @@ class Tarmac(val parent: Configurable) extends QScript with PedigreeQscript with
...
+     * Create XHMM count file or create link to input count file
+     * Precedence is given to existing count files.
+     * Returns a disjunction where right is the file, and left is
+     * a potential error message
+     */
+    protected lazy val outputXhmmCountJob: String \/ QFunction = {
+      val outFile = new File(sampleDir + File.separator + s"$name.dcov")
+      (inputXhmmCountFile, bamFile) match {
+        case (Some(f), _) => {
+          val ln = new Ln(root)
```
ffinfo: Maybe add here something like this:
`if (bamFile.isDefined) logger.warn(s"bam file and count file are both given for $sampleId, bam file is ignored")`

ffinfo (changes requested):
Only a small comment on the bamFile, de rest is oke for me

ffinfo (approved):

PR #55, opened 2017-04-04
Fix biopet 306

akaljuvee:


akaljuvee:
hopefully one of you has time to look at this

shiva/src/test/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingTest.scala
```diff
@@ -90,7 +90,7 @@ class ShivaSvCallingTest extends TestNGSuite with Matchers {
       pipeline.init()
       pipeline.script()
 
-      val summaryCallers = pipeline.summarySettings("sv_callers")
+      val summaryCallers = pipeline.summarySettings.get("sv_callers").get.asInstanceOf[List[String]]
```
ffinfo, line 93: Generally in scala it's not advised to use .get on a option (not confuse this with .get on a Map), this can result in a None.get error.
Also doing asInstanceOf is not always very safe. This is only safe when you first check the type. In scala you would write is like this:
``` scala
val summaryCallers: List[String] = pipeline.summarySettings.get("sv_callers") match {
  case Some(callers: List[String]) => callers
  case None => throw new Exeption("'sv_callers' should exist in summarySettings")
  case _ => throw new Exeption("'sv_callers' should be a List[String]")
}
```

Smaller but less safe, gives an key not found exception when sv_callers is not found:
``` scala
val summaryCallers: List[String] = pipeline.summarySettings("sv_callers") match {
  case callers: List[String] => callers
  case _ => throw new Exeption("'sv_callers' should be a List[String]")
}
```

This is test code so it's less an issue but still good do it like this as good practice.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
```diff
@@ -0,0 +1,110 @@
...
+    if (dupCounts.exists(elem => (elem._2.sum > 0))) result += ("DUP" -> dupCounts)
+    if (invCounts.exists(elem => (elem._2.sum > 0))) result += ("INV" -> invCounts)
+    result
+  }
+
+  def parseSummaryForTranslocations(summary: SummaryDb, runId: Int, sampleNames: Seq[String]): Map[String, Long] = {
+    var traCounts: Map[String, Long] = Map()
+    for (sampleName <- sampleNames) {
+      val counts: Map[String, Any] = Await.result(summary.getStat(runId, PipelineName("shivasvcalling"), ModuleName("parse_sv_vcf"), SampleName(sampleName)), Duration.Inf).get
+      traCounts += (sampleName -> counts.get("TRA").get.asInstanceOf[Long])
```
ffinfo: please don't use .get and .asInstanceOf in production code. Only use this when there is really no other way. Please replace this with match statements.

shiva/src/test/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingTest.scala
```diff
@@ -182,7 +182,7 @@ class ShivaSvCallingTest extends TestNGSuite with Matchers {
     pipeline.init()
     pipeline.script()
 
-    val summaryCallers = pipeline.summarySettings("sv_callers")
+    val summaryCallers: List[String] = pipeline.summarySettings.get("sv_callers").get.asInstanceOf[List[String]]
```
ffinfo, line 185: same as above

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -96,17 +97,54 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
...
-  def summarySettings = Map("sv_callers" -> configCallers.toList)
+  def summarySettings = Map("sv_callers" -> configCallers.toList, "hist_bin_boundaries" -> ShivaSvCallingReport.histogramBinBoundaries)
 
   /** Files for the summary */
-  def summaryFiles: Map[String, File] = Map("final_mergedvcf" -> (if (inputBams.size > 1) outputMergedVCF else outputMergedVCFbySample.values.head))
+  def summaryFiles: Map[String, File] = if (inputBams.size > 1) Map("final_mergedvcf" -> outputMergedVCF) else Map.empty
+
+  def getVariantCounts(vcfFile: File, breaks: Array[Int]): Map[String, Any] = {
+    val delCounts, insCounts, dupCounts, invCounts = Array.fill(breaks.size + 1) { 0 }
+    var traCount = 0
```
bow: What is the reason of using `var Int` for `traCount` but `val Array[Int]` for the other SV types?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -96,17 +97,54 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
...
+
+    val iterator = new VCFFileReader(vcfFile, false).iterator
```
bow: I think a `.foreach` on the iterator may also work here. I think (not 100% sure) that this means you can also skip closing the iterator explicitly afterwards.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -96,17 +97,54 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
...
+    while (iterator.hasNext) {
+      val record = iterator.next
+      val svType = record.getAttributeAsString("SVTYPE", "")
```
bow: You can also match directly here, e.g.:

```
record.getAttributeAsString("SVTYPE", "") match {
    case "TRA" | "CTX" | "ITX" => traCount += 1
    case otherwise => {
         ....
    }
}
```

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  val histogramBinBoundaries: Array[Int] = Array(100, 1000, 10000, 100000, 1000000, 10000000)
+  val histogramPlotTicks: Array[Int] = Array(100, 1000, 10000, 100000, 1000000, 10000000, 100000000)
+  val histogramText: List[String] = List("<=100bp", "0.1-1kb", "1-10kb", "10-100kb", "0.1-1Mb", "1-10Mb", ">10Mb")
+
+  def parseSummaryForSvCounts(summary: SummaryDb, runId: Int, sampleNames: Seq[String]): Map[String, Map[String, Array[Long]]] = {
+    var delCounts, insCounts, dupCounts, invCounts: Map[String, Array[Long]] = Map()
+
+    for (sampleName <- sampleNames) {
+      val sampleCounts: Map[String, Any] = Await.result(summary.getStat(runId, PipelineName("shivasvcalling"), ModuleName("parse_sv_vcf"), SampleName(sampleName)), Duration.Inf).get
```
bow: Would it be possible to, instead of doing `Await.result` here, make the `parseSummaryForSvCounts` return a `Future`instead? So it's the caller of this function that will need to do the `Await`.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
```diff
@@ -0,0 +1,110 @@
...
+
+    tsvWriter.print("histogramBin")
+    sampleNames.foreach(sampleName => tsvWriter.print("\t" + sampleName))
+    tsvWriter.println()
+
+    for (i <- histogramPlotTicks.indices) {
+      tsvWriter.print(histogramPlotTicks(i))
+      sampleNames.foreach(sampleName => tsvWriter.print("\t" + counts.get(sampleName).get(i)))
+      tsvWriter.println()
+    }
```
bow: You can also iterate over `histogramPlotTicks` while getting each item's index. Something like:

```
histogramPlotTicks.zipWithIndex.foreach {
    case (item, index) => {
        tsvWriter.print(item)
        sampleNames.foreach(sampleName => tsvWriter.print("\t" + counts.get(sampleName).get(index))
        tsvWriter.println()
    }
}
```

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -96,17 +97,54 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
     // group by "tags"
     // sample tagging is however not available within this pipeline
 
+    for ((sample, mergedResultFile) <- outputMergedVCFbySample) {
+      lazy val counts = getVariantCounts(mergedResultFile, ShivaSvCallingReport.histogramBinBoundaries)
```
ffinfo: This means the WriteSummary job is parsing a vcf file. This is a to high cpu load because this is not running on a compute node. This should move to a tool that we can submit. Vcfstats already generate some stats from a vcf file so it would be good to add this to this tool.

bow (dismissed):
Hey @akaljuvee,

I made some comments on parts of the code. I'll leave parts that deal more with the reporting template to others who understand them more :).

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -96,17 +97,54 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
...
   /** Files for the summary */
-  def summaryFiles: Map[String, File] = Map("final_mergedvcf" -> (if (inputBams.size > 1) outputMergedVCF else outputMergedVCFbySample.values.head))
+  def summaryFiles: Map[String, File] = if (inputBams.size > 1) Map("final_mergedvcf" -> outputMergedVCF) else Map.empty
+
+  def getVariantCounts(vcfFile: File, breaks: Array[Int]): Map[String, Any] = {
+    val delCounts, insCounts, dupCounts, invCounts = Array.fill(breaks.size + 1) { 0 }
+    var traCount = 0
+
+    val iterator = new VCFFileReader(vcfFile, false).iterator
+    while (iterator.hasNext) {
```
ffinfo: This is really a java way of iterator. Better use 
``` scala
import scala.collection.JavaConversions._ // this should be add the top

for (record <- reader) {
}
```

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
ffinfo: The reader should be in an other val so you can close it.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
ffinfo: No need to make use of vars here, a for/map statement can return a value

shiva/src/main/resources/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariantsSv.ssp
```diff
@@ -0,0 +1,97 @@
...
+}#
+
+
+#if (showPlot)
+    <div class="panel-body">
+    #for (sv <- svTypes)
+        <img src="${sv.pngFileName}" />
+    #end
+    </div>
+    #if (traCounts.nonEmpty)
```
ffinfo, line 42: This should be always displayed even when there are no translocations found

shiva/src/main/resources/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariantsSv.ssp
```diff
@@ -0,0 +1,97 @@
...
+        <h5>Number of translocation events detected:</h5>
+        <table class="table table-condensed" style="width:auto">
+            <thead><tr>
+            #for (sampleName <- sampleNames)
+                <th>${sampleName}</th>
+            #end
+            </tr></thead>
+            <tbody><tr>
+            #for (sampleName <- sampleNames)
+                <td>${traCounts.get(sampleName).get}</td>
```
ffinfo: `traCounts.get(sampleName).get` is the same as `traCounts(sampleName)`.

Also it's safer to use it like this: `traCounts.getOrElse(sampleName, 0)`

shiva/src/main/resources/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariantsSv.ssp
```diff
@@ -0,0 +1,97 @@
...
+        </button>
+        <a href="${tsvAllTypes}"><button type="button" class="btn btn-info"><i class="glyphicon glyphicon-cloud-download"></i> TSV file</button></a>
+    </div>
+#end
+
+<div class="panel-body collapse #if (showTable)in#end" id="svSummaryTable">
+
+#for (sv <- svTypes)
+#{
+    val countsForSvType: Map[String, Array[Long]] = counts.get(sv.svType).get
```
ffinfo: same as above

shiva/src/main/resources/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariantsSv.ssp
```diff
@@ -0,0 +1,97 @@
...
+    <thead><tr><th data-sorted="true" data-sorted-direction="ascending">Sample</th>
+        #for (text <- ShivaSvCallingReport.histogramText)
+            <th>${text}</th>
+         #end
+    </tr></thead>
+    <tbody>
+        #for (sampleName <- sampleNames)
+            <tr>
+            <td><a href="${rootPath}Samples/${sampleName}/index.html">${sampleName}</a></td>
+            #for (countForSize <- countsForSvType.get(sampleName).get)
```
ffinfo: same as above

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
```diff
@@ -48,13 +48,22 @@ trait ShivaReportTrait extends MultisampleMappingReportTrait {
...
 
   override def pipelineName = "shiva"
 
-  override def additionalSections = super.additionalSections ++ (if (variantcallingExecuted) List("Variantcalling" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariants.ssp",
-    Map("showPlot" -> true, "showTable" -> false)))
-  else Nil)
+  override def additionalSections = {
+    val params = Map("showPlot" -> true, "showTable" -> false)
+    super.additionalSections ++
+      (if (variantcallingExecuted) List("SNV Calling" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariants.ssp", params)) else Nil) ++
```
ffinfo, line 64: Calling is SNV Calling suggest only single nucleotide changes are found while there is also small indels in normal variantcalling

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
```diff
@@ -107,9 +116,10 @@ trait ShivaReportTrait extends MultisampleMappingReportTrait {
 
   /** Single sample page */
   override def samplePage(sampleId: Int, args: Map[String, Any]): ReportPage = {
-    val variantcallingSection = if (variantcallingExecuted) List("Variantcalling" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariants.ssp")) else Nil
+    val variantcallingSection = if (variantcallingExecuted) List("SNV Calling" -> ReportSection("/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariants.ssp")) else Nil
```
ffinfo: same as above

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
```diff
@@ -0,0 +1,110 @@
...
+          case "DEL" => delCounts += elem
+          case "INS" => insCounts += elem
+          case "DUP" => dupCounts += elem
+          case "INV" => invCounts += elem
+        }
+      }
+    }
+
+    var result: Map[String, Map[String, Array[Long]]] = Map()
+    if (delCounts.exists(elem => (elem._2.sum > 0))) result = Map("DEL" -> delCounts)
```
ffinfo, line 39: Does this means that if there is nothing found it also does not report that there are 0 deletions found?

ffinfo (changes requested):

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
akaljuvee: I figured that it's better not to show anything for a particular sv type when Biopet hasn't been configured to detect it. Take for example inversions. With the default configuration these will never be reported, by default all tools will be run (except for pindel), Pysvtools is used for merging and as one of the tools reports only dels and ins, then no inversions in the output. I considered it incorrect to report then, that in the sample being analyzed there were no inversions found.

Which sv types are shown could be decided based on Shiva's conf, but for that I'd need to know more details about what this Pysvtools is doing. I remember from testing that it can for example combine variants that have been marked with different svtype tag in the vcf-file, for example combining insertions and duplications. So I skipped studying Pysvtools and used a simpler approach, assuming that when one particular sv type wasn't detected in any of the samples, then one of the tools didn't report that type and then it can't be in the output.

You think the zeros should be present there all the time? Or it's worth writing code to make this decision based on configuration?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
akaljuvee: I don't know the size of the translocated fragments. For each translocation there are 2 positions given in the vcf file, one from each chromosome affected (in case of interchromosomal tra.-s), I don't have info about the size.

ffinfo: Most of this tools detect only breakpoints, this mean that for each SV can takes more then 1 breakpoint to detect the size is not informative. Only deletions and insertions can be called with a single breakpoint (or single call).

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
ffinfo: Let's not do this based on config values. We don't know if a tools can't detect some variant types or there are simply none found. Also, maybe some tools add some functionality in the future.

akaljuvee: We do know which sv types each method is designed to detect, but I didn't do this based on conf because I don't know the details about how Pysvtools is combining the results.
For me it was more logical not to show the rows with zeros, but if you prefer the other way, then I can make the change.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
bow: Aha, I get it now. So `breaks` is an array of counts per SV size? If so, this then relies on `breaks` having a specific ordering of boundaries (that is not fully apparent at this point in the code).

What about specifically encoding the SV size boundaries? Something like this:

```
case class SvBoundary(lower: Long, upper: Long) {
    def containsSize(svSize: Long) = lower <= svSize && upper >= svSize
}
```

You would then use this case class's `containsSize` to match a given SV size.

The counts can then be stored in something like `collection.mutable.Map[SvBoundary, Int]`.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
sndrtj: I do agree with @akaljuvee that, in the case we are certain we cannot detect a certain class of variants, we should not report those variant as zero. There is a semantic difference between "no variants found" and "cannot technically detect this class of variants".  In the latter case, you don't know whether you have any variants of this type.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
akaljuvee: 'breaks' is the param for histogram bin boundaries. I agree, that the name doesn't help with understanding what's going on. It came from R's histogram, there the boundaries are called the same way. So yes, smth should be changed when it's not clear how the sv-s are counted. But maybe here simply adding comments to the code would be enough? I could also remove this 'breaks' parameter and use the static value ShivaSvCallingReport.histogramBinBoundaries. Then noone would access the method with incorrect input. In ShivaSvCallingReport there's also the val histogramText, seeing that should also make it more clear.

bow: Ok. Using a static is preferrable, along with the documentation.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
akaljuvee: I don't see what would be the benefit of doing that. ssp-files are rendered by the scalate framework and I don't believe it's possible to propagate this Future to the processing that library is doing.

ffinfo: This is indeed true. In the future we need to look into to see if we can make this template more static, so no SummaryDb calls in there anymore. For now this can stay like this.

bow: Aha, ok. I was under the impression that the values gets called from outside the template first before they get passed on (i.e. all `Future`s are first waited on). That's where I expect the Future to be awaited.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
akaljuvee: I don't understand here what you mean by this

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCallingReport.scala
akaljuvee: @ffinfo, this small question is also open.

ffinfo: The problem is here maintaining this. On each change of a tool and when we add more tools we need to see if something did change. This is not something we want to spend time on.

All we can do is put a disclaimer in the report that 0 can mean that it's not found or not detectable but that is something up to the tool.

ffinfo (approved):

PR #58, opened 2017-04-06
Option to skip printreads and give bqsr files to the variantcalller

ffinfo:
This will fix BIOPET-635.

With this is becomes possible to skip printreads (this step is using a lot of diskspace) and suply the bqsr files to the variantcaller.

Please note that this is only working for GATK callers.


shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
```diff
@@ -131,6 +146,41 @@ class Shiva(val parent: Configurable) extends QScript with MultisampleMappingTra
...
+        add(baseRecalibrator)
+
+        if (useAnalyzeCovariates) {
+          val baseRecalibratorAfter = BaseRecalibrator(qscript, inputBam, bqsrAfterFile.get)
+          baseRecalibratorAfter.BQSR = bqsrFile
+          add(baseRecalibratorAfter)
+          add(AnalyzeCovariates(qscript, baseRecalibrator.out, baseRecalibratorAfter.out, swapExt(dir, inputBam, ".bam", ".baserecal.pdf")))
+        }
+
+        val printReads = PrintReads(qscript, inputBam, swapExt(dir, inputBam, ".bam", ".baserecal.bam"))
```
sndrtj: Lines 164-166 could be in the `usePrintreads` if block.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
```diff
@@ -131,6 +146,41 @@ class Shiva(val parent: Configurable) extends QScript with MultisampleMappingTra
           add(vc)
         })
       }
+
+      /** Adds base recalibration jobs */
+      def addBaseRecalibrator(inputBam: File, dir: File, isIntermediate: Boolean, usePrintreads: Boolean): File = {
+        val baseRecalibrator = BaseRecalibrator(qscript, inputBam, bqsrFile.get) // at this point bqsrFile should exist
```
sndrtj: Explicitly checking whether the `bqsrFile` is defined may not be a bad thing, especially as this is a function that could be called before it is defined at some point in the future.

bow (approved):

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
ffinfo: Build in a other check

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
ffinfo: agree, this was still there for old code ;)

PR #59, opened 2017-04-06
Tarmac reference building

sndrtj:


tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -33,7 +53,93 @@ class Tarmac(val parent: Configurable) extends QScript with PedigreeQscript with
   }
 
   def addMultiSampleJobs() = {
+    val initRefMap = samples map { x => x._2 -> getReferenceSamplesForSample(x._1) }
```
bow: to make this more readable, you can also pattern match on the tuple directly:
```
val initRefMap = samples.map { case (a, b) => b -> getReferenceSamplesForSample(a) }
```
(of course `a` and `b` should be named appropriately).

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -33,7 +53,93 @@ class Tarmac(val parent: Configurable) extends QScript with PedigreeQscript with
...
+    initRefMap.values.filter(_.isLeft).foreach {
+      case -\/(message) => Logging.addError(message)
+      case _            => ;
+    }
```
bow: You can shorten this like so:
```
initRefMap.values.collect { case -\/(err) => err }.foreach { Logging.addError }
```

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -33,7 +53,93 @@ class Tarmac(val parent: Configurable) extends QScript with PedigreeQscript with
...
+
+    val refMap: Map[Sample, Set[Sample]] = initRefMap.filter(_._2.isRight).map { x =>
```
bow: Similar to the above, I think using `.collect` and pattern matching would make this more clear.

bow (changes requested):
Hey @sndrtj,

Looks ok in general, but I see several places that `\/` is filtered then on the subsequent map it is extracted manually. This can be combined to a single operation using `.collect`.

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
```diff
@@ -33,7 +53,93 @@ class Tarmac(val parent: Configurable) extends QScript with PedigreeQscript with
   }
 
   def addMultiSampleJobs() = {
+    val initRefMap = samples map { x => x._2 -> getReferenceSamplesForSample(x._1) }
+    initRefMap.values.filter(_.isLeft).foreach {
+      case -\/(message) => Logging.addError(message)
+      case _            => ;
```
ffinfo: `;` is redundant

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
ffinfo: `.map` works here but maybe do is like this:
``` scala
.map { case (key, value) => 
  // your code
}
```

ffinfo (changes requested):

ffinfo:
I think all test should be unit tested, is there a way to do this already on this branch?

tarmac/src/main/scala/nl/lumc/sasc/biopet/pipelines/tarmac/Tarmac.scala
sndrtj: Compiler gives a warning message if not.

ffinfo (approved):

PR #62, opened 2017-04-07
Fixing fastqc plots

ffinfo:
(cherry picked from commit b5863e4)


ffinfo:
This is failing but the fix is already in https://github.com/biopet/biopet/pull/57

rhpvorderman (commented):
If test builds complete is should be okay.

rhpvorderman (approved):
If it builds it should be okay

PR #65, opened 2017-04-10
Fixing nullpointer

ffinfo:


biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetCommandLineFunction.scala
```diff
@@ -99,7 +99,7 @@ trait BiopetCommandLineFunction extends CommandLineResources { biopetFunction =>
     beforeGraph()
     internalBeforeGraph()
 
-    this.commandDirectory = this.jobOutputFile.getParentFile
+    if (jobOutputFile != null) this.commandDirectory = this.jobOutputFile.getAbsoluteFile.getParentFile
```
akaljuvee, line 103: the reason why 'jobOutputFile' is missing is not important?

akaljuvee (approved):

PR #68, opened 2017-04-12
Fixing report failures

ffinfo:


rhpvorderman (approved):
The code has also become much more readable in the process.

PR #81, opened 2017-04-18
Added gcDistribution function.

rhpvorderman:


flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/Fastqc.scala
```diff
@@ -86,6 +86,15 @@ class Fastqc(root: Configurable) extends nl.lumc.sasc.biopet.extensions.Fastqc(r
     else fqModules
   }
 
+  def gcDistribution: Option[Map[Int, Double]] = {
```
ffinfo: Could you add some scaladocs here?

ffinfo (changes requested):

flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/Fastqc.scala
rhpvorderman: It is now added.

ffinfo (approved):

PR #88, opened 2017-04-20
Corrections on threads

ffinfo:


biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/CommandLineResources.scala
```diff
@@ -22,6 +22,8 @@ import org.broadinstitute.gatk.queue.function.CommandLineFunction
  */
 trait CommandLineResources extends CommandLineFunction with Configurable {
 
+  lazy val maxThreads: Option[Int] = config("max_threads")
+
   def defaultThreads = 1
```
akaljuvee: what is config("threads") and config("max_threads")? would be better if it was explained here with code comments

ffinfo: Agree, will fix this

akaljuvee (approved):

PR #89, opened 2017-04-21
fixes for sv calling

akaljuvee:


shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaSvCalling.scala
```diff
@@ -43,7 +48,16 @@ class ShivaSvCalling(val parent: Configurable) extends QScript with SummaryQScri
 
   /** Executed before script */
   def init(): Unit = {
-    if (inputBamsArg.nonEmpty) inputBams = BamUtils.sampleBamMap(inputBamsArg)
+    if (inputBamsArg.nonEmpty) {
+      inputBams = BamUtils.sampleBamMap(inputBamsArg)
+
+      val db = SummaryDb.openSqliteSummary(summaryDbFile)
+      for (sampleName <- inputBams.keys) {
+        if (Await.result(db.getSampleId(summaryRunId, sampleName), Duration.Inf).isEmpty) {
```
ffinfo, line 56: It might be better to first do a query to fetch all samples for the current run instead of a single query for each sample. For now this works so no need to change this.

ffinfo (approved):

PR #90, opened 2017-04-25
Fixing small bug for coming release

ffinfo:


gears/src/main/scala/nl/lumc/sasc/biopet/pipelines/gears/GearsSingle.scala
```diff
@@ -55,6 +60,13 @@ class GearsSingle(val parent: Configurable) extends QScript with SummaryQScript
     if (fastqR2.nonEmpty && fastqR1.size != fastqR2.size) Logging.addError("R1 and R2 has not the same number of files")
     if (sampleId == null || sampleId == None) Logging.addError("Missing sample ID on GearsSingle module")
 
+    if (!skipFlexiprep) {
+      val db = SummaryDb.openSqliteSummary(summaryDbFile)
```
bow, line 64: how about using:

```
val .... = for {
    sample <- SummaryDb.openSqliteSummary(summaryDbFile)
    sId <- sample.map(...)
    library <- ...
    libId <- ...
} yield (...)
```

If you need to keep the `Future` without unwrapping its contents, you can also use `=` instead of `<-`.

rhpvorderman (approved):

PR #96, opened 2017-04-26
Add documentation for freec

johnmous:


docs/pipelines/kopisu.md
```diff
@@ -37,7 +37,35 @@ All three methods can be run concurrently. However, we do not yet provide any me
...
+  reference_name: GRCh38_no_alt_analysis_set
+  reference_fasta: /exports/genomes/species/H.sapiens/GRCh38_no_alt_analysis_set/reference.fa
+  chrFiles: /path/to/referenceFastaFiles/
+  chrLenFile: /path/to/referenceIndex/reference.fa.fai
+  coefficientOfVariation: 0.05
+  output_dir: /path/to/output
+  ploidy: 2
+  mateOrientation: 0
+  inputFormat: BAM
+  bedtools:
```
ffinfo: I think it might be better to note it like lhis:
``` yaml
freec:
  exe: <freec executable>
  bedtools:
    exe: <executable bedtools>
```

You should test this but it was intended to work like this.

johnmous: If I use the above, I get this into the freec config file:  `bedtools=Map(exe` -> /usr/local/bin/bedtools)`
It does not work when bedtools is under freec

docs/pipelines/kopisu.md
```diff
@@ -48,25 +48,39 @@ Freec is a tool that needs a few tricks and workarounds to make it work. The wor
 3. `mateOrientation` should be set to 0. 
 
 **Freec config example**
```
ffinfo: I think this title is duplicated at the wrong place?

docs/pipelines/kopisu.md
ffinfo: oke sure, if this does work it's also fine:
``` yaml
freec:
  exe: <freec executable>
bedtools:
  exe: <executable bedtools>
```

docs/pipelines/kopisu.md
```diff
@@ -37,7 +37,49 @@ All three methods can be run concurrently. However, we do not yet provide any me
...
+  use_freec_method: true  
+  output_dir: /exports/sasc/project-200-Karoly-SV/analysis/04_kopisu
+  reference_name: GRCh38_no_alt_analysis_set
+  reference_fasta: /exports/genomes/species/H.sapiens/GRCh38_no_alt_analysis_set/reference.fa
+  chrFiles: /exports/sasc/project-200-Karoly-SV/analysis/03_kopisu/freec_method/testFREEC/reference/
+  chrLenFile: /exports/sasc/project-200-Karoly-SV/analysis/03_kopisu/freec_method/testFREEC/reference.fa.fai
+  coefficientOfVariation: 0.05
+  ploidy: 2
+  mateOrientation: 0
+  inputFormat: BAM
```
ffinfo: The input from kopisu is always a bam file. Still this option is required? if so we should add this to the defaults of kopisu itself so the user does not need to supply them.

docs/pipelines/kopisu.md
```diff
@@ -37,7 +37,49 @@ All three methods can be run concurrently. However, we do not yet provide any me
...
+**Freec config example**
+
+```
+kopisu: 
+  use_freec_method: true  
+  output_dir: /exports/sasc/project-200-Karoly-SV/analysis/04_kopisu
+  reference_name: GRCh38_no_alt_analysis_set
+  reference_fasta: /exports/genomes/species/H.sapiens/GRCh38_no_alt_analysis_set/reference.fa
+  chrFiles: /exports/sasc/project-200-Karoly-SV/analysis/03_kopisu/freec_method/testFREEC/reference/
+  chrLenFile: /exports/sasc/project-200-Karoly-SV/analysis/03_kopisu/freec_method/testFREEC/reference.fa.fai
```
ffinfo: This fai file can be retrieved inside the pipeline, if this value need to be set we should add this to the defaults of kopisu

docs/pipelines/kopisu.md
```diff
@@ -37,7 +37,49 @@ All three methods can be run concurrently. However, we do not yet provide any me
...
+
+3. `mateOrientation` should be set to 0. 
+
+
+**Freec config example**
+
+```
+kopisu: 
+  use_freec_method: true  
+  output_dir: /exports/sasc/project-200-Karoly-SV/analysis/04_kopisu
```
ffinfo: Could you remove the real output dirs and replace them be fake ones or placeholders? This documentation is public.

docs/pipelines/kopisu.md
johnmous: Yes, the inputFormat is required, as long as we provide the mateFile (alignment file). I do not think there is any use of FREEC without the alignment file for us. So you can set inputFormat to BAM as default

docs/pipelines/kopisu.md
johnmous: Actually, the user must make a copy of the fai file and edit in the case chromosome names do not start with "chr".  "chr" should bee added manually. This is a known issue: https://github.com/BoevaLab/FREEC/issues/2#issuecomment-238261681

sndrtj (dismissed):
Three paragraphs are repeated ad verbatim. Please remove one of the copies. 

Also, I think it would be really nice if you could give an example filesystem tree of a reference (with say, 3 contigs), for extra clarity.

johnmous:
Removed the duplicated piece. And added a filesystem tree with 3 example fasta files.

ffinfo (approved):

PR #98, opened 2017-04-28
Adding pipeline for gennotyping gvcf files

ffinfo:


shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/GenotypeGvcfs.scala
```diff
@@ -0,0 +1,78 @@
...
+    val combineJob = new CombineJob(finalGvcfFile, outputDir, inputGvcfs)
+    combineJob.allJobs.foreach(add(_))
+
+    val genotype = new gatk.GenotypeGVCFs(this)
+    genotype.variant = List(finalGvcfFile)
+    genotype.out = finalVcfFile
+    add(genotype)
+  }
+
+  class CombineJob(outputFile: File, outputDir: File, allInput: List[File], group: List[Int] = Nil) {
```
akaljuvee, line 47: why do you need a class for this?

akaljuvee: For me the constructor of a class is simply for creating it, setting the instance variables and that's it. The rest goes into methods. Is there a reason why this code has to be inside the constructor? I'd move it to a recursive method instead.

akaljuvee (approved):

PR #106, opened 2017-05-04
Fixing input files for genotyping

ffinfo:


shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/HaplotypeCallerGvcf.scala
```diff
@@ -124,6 +124,7 @@ class HaplotypeCallerGvcf(val parent: Configurable) extends Variantcaller {
     }
 
     genotypeGvcfs.outputDir = outputDir
+    genotypeGvcfs.namePrefix = s"$namePrefix.$name"
```
rhpvorderman, line 127: namePrefix is defined as prefix+name ? Wouldn't the variable than better be called genotpeGcvfs.nameWithPrefix?

rhpvorderman (changes requested):

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/HaplotypeCallerGvcf.scala
ffinfo: Not directly. This name prefix is for genotypeGvcfs, the other is from HaplotypecallerGvcf. Only in HaolotpecallerGvcf I do set the namePrefix of GenotypeGvcfs

rhpvorderman (approved):

PR #107, opened 2017-05-05
Added annotation files in gentrap report

rhpvorderman:


ffinfo (approved):
This oke for now. Later we also need to check the cufflinks modules, that's also using gtf files.

PR #110, opened 2017-05-09
Feature taxextract

sndrtj:


docs/pipelines/tinycap.md
```diff
@@ -72,6 +72,32 @@ One can specify other options such as: `bowtie` (alignment) options, clipping an
 }
 ```
 
+## Taxonomy extraction
```
ffinfo, line 75: I see a lot of duplication here. Maybe better to make a separated page and link to that from all those pages?

gears/src/main/scala/nl/lumc/sasc/biopet/pipelines/gears/GearsSingle.scala
```diff
@@ -115,19 +158,17 @@ class GearsSingle(val parent: Configurable)
 
   protected def executeFlexiprep(r1: List[File], r2: List[File]): (File, Option[File]) = {
     val read1: File =
-      if (r1.size == 1) r1.head
+      if (r1.size == 1) mergedR1
       else {
-        val outputFile = new File(outputDir, "merged.R1.fq.gz")
-        add(Zcat(this, r1) | new Gzip(this) > outputFile)
-        outputFile
+        add(Zcat(this, r1) | new Gzip(this) > mergedR1)
```
ffinfo: Can you make this intermediate?

gears/src/main/scala/nl/lumc/sasc/biopet/pipelines/gears/GearsSingle.scala
```diff
@@ -115,19 +158,17 @@ class GearsSingle(val parent: Configurable)
...
       }
 
     val read2: Option[File] =
-      if (r2.size <= 1) r2.headOption
+      if (r2.size <= 1) mergedR2
       else {
-        val outputFile = new File(outputDir, "merged.R2.fq.gz")
-        add(Zcat(this, r2) | new Gzip(this) > outputFile)
-        Some(outputFile)
+        add(Zcat(this, r2) | new Gzip(this) > mergedR2.get)
```
ffinfo: Can you make this intermediate?

mapping/src/main/scala/nl/lumc/sasc/biopet/pipelines/mapping/MultisampleMapping.scala
```diff
@@ -88,6 +88,24 @@ trait MultisampleMappingTrait extends MultiSampleQScript with Reference { qscrip
   def makeSample(id: String) = new Sample(id)
   class Sample(sampleId: String) extends AbstractSample(sampleId) { sample =>
 
+    val gearsJob: Option[GearsSingle] = mappingToGears match {
```
ffinfo: The name makes me think it's a single job but this is a subpipeline

docs/pipelines/tinycap.md
```diff
@@ -72,6 +72,32 @@ One can specify other options such as: `bowtie` (alignment) options, clipping an
...
+
+Taxonomies are determined using [Gears](gears.md) as a sub-pipeline. 
+
+To enable taxonomy extraction, specify the following additional flags in your
+config file:
+
+| Name | Namespace | Type | Function |
+| ---- | --------- | ---- | -------- |
+| mapping_to_gears | mapping | Boolean | Must be set to **true** |
+| taxonomy_extract | mapping | Boolean (must be **true** for this purpose) | enable taxonomy extraction |
```
ffinfo, line 91: Maybe a example config can help here?

ffinfo (changes requested):
In general looking great. Just a few comments.

Also, please fix the failing test ;).

docs/pipelines/tinycap.md
sndrtj: I understand your point. However, there are some reasons why I think that may not be the best approach:

1. To understand all of the options of a pipeline, the user already has to click through several pages. Adding yet another would place yet more cognitive burden. Having all options pertaining to one pipeline in one document may look a bit cleaner.
2. Where to put such a stand-alone taxonomy extraction document is also not entirely clear. It's not a pipeline, so it wouldn't belong under `pipelines`. It's also not generalizable to _all_ pipelines, so placing it under `general` also feels wrong. 

That said, I do certainly understand the duplication isn't the best from a maintenance POV. Preferably, I'd like to import the section in all relevant pipelines. I'll look if such a thing is possible.

ffinfo: Maybe your right for now indeed. What we should have is a MultiSampleMapping page that explaining all the common features. Maybe we should make a structure like this:
```
- pipelines
  - multisamplemapping
    - Shiva
    - Gentrap
    - Carp
  - Other pipelines
```

Will make a new issue to do this in the future

ffinfo (approved):

PR #113, opened 2017-05-16
Fixed bug when python file already exist

ffinfo:


sndrtj:
Hi, 

This will now fail when `setPythonScript` is called with an absolute path to a pre-existing script. The failing tests on Kopisu are an example of that.

PR #116, opened 2017-05-23
Adding validation step for annotation files in Gentrap

ffinfo:


biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
```diff
@@ -0,0 +1,141 @@
...
+            .getOrElse(cdsFeatures.getOrElse(Nil)) ::: stopFeatures.getOrElse(
+            cdsFeatures.getOrElse(Nil))
+          val codingStart =
+            if (codingLocations.isEmpty) None
+            else if (gtfGene.strand.get) Some(codingLocations.min - 1)
+            else Some(codingLocations.min - 1)
+          val codingEnd =
+            if (codingLocations.isEmpty) None
+            else if (gtfGene.strand.get) Some(codingLocations.max)
+            else Some(codingLocations.max)
```
akaljuvee: there's a bug here, with these min-s and max

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
```diff
@@ -0,0 +1,141 @@
...
+          Transcript(transcriptId.get,
+                     gtfTranscript.start,
+                     gtfTranscript.end,
+                     codingStart.getOrElse(gtfTranscript.end),
+                     codingEnd.getOrElse(gtfTranscript.end),
```
akaljuvee, line 100: and a small bug here

gentrap/src/main/scala/nl/lumc/sasc/biopet/pipelines/gentrap/Gentrap.scala
```diff
@@ -136,6 +136,20 @@ class Gentrap(val parent: Configurable)
     )
   )
 
+  override def biopetScript(): Unit = {
+    val validate = new ValidateAnnotation(this)
+    validate.refflatFile = annotationRefFlat.get
+    validate.gtfFile = fragmentsPerGene.map(_.annotationGtf)
```
akaljuvee: * this makes having the annotation in refflat obligatory, why so?
* cufflinks also uses the gtf format

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
ffinfo: Nope, this is on purpose. When no coding regions on a transcript the start and end of the coding will be the end of the transcript.

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
ffinfo: can you explain the bug because I can't see it. Also I did validate the results to a known set, the results are almost 100% the same, except where the initial tool was wrong.

gentrap/src/main/scala/nl/lumc/sasc/biopet/pipelines/gentrap/Gentrap.scala
ffinfo: Nope, gtfFile is a optional argument. The tool also works with only a refflat file.
Cufflinks your right about that, will see if I can put this somewhere in.

ffinfo: btw, a refflat file is indeed a required option to gentrap

akaljuvee: why? when I want only to use cufflinks, then I don't need the annotation to be in refflat format

ffinfo: In gentrap there is always picard CollectRnaSeqMetrics executed, this requires a refflat file

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
akaljuvee: when the transcript is on + strand, then for getting the start coordinate you should use codingLocations.min, when it's on - strand, then you should use codingLocations.max, isn't it so?

ffinfo: well this is where refflat and gtf files are different. In a refflat file are coding region relative to the reference. So for a negative transcript coding start in a refflat file is the end of the transcript. I know this is confusing ;)

akaljuvee: for what is this 'else if' there, you only need else for the same result

ffinfo: Good question. This is because in a gtf file there are 2 ways of defining coding regions. They can use the feature `CDS` or a combination of `start_codon` and `stop_codon`. In annotation files this can be used both but not consistently :(.

akaljuvee: I simply meant that
val codingStart =
if (codingLocations.isEmpty) None
else if (gtfGene.strand.get) Some(codingLocations.min - 1)
else Some(codingLocations.min - 1)

is the same as
val codingStart =
if (codingLocations.isEmpty) None
else Some(codingLocations.min - 1)

don't understand why there's the check for strand if you don't use it for the resulting value

ffinfo: Ohh, your right, fixed that now :)

gentrap/src/main/scala/nl/lumc/sasc/biopet/pipelines/gentrap/CheckValidateAnnotation.scala
```diff
@@ -0,0 +1,47 @@
...
+  @Input(required = true)
+  var inputLogFile: File = _
+
+  /** Exits whenever the input md5sum is not the same as the output md5sum */
+  def run: Unit = {
+
+    val reader = Source.fromFile(inputLogFile)
+    reader.getLines().foreach { line =>
+      if (line.startsWith("ERROR")) {
+        logger.error("Corrupt vcf file found, aborting pipeline")
```
akaljuvee: someone could get confused seeing the error is about vcf files

ffinfo: Also did see that just now, is fixed now

akaljuvee (approved):
would be good if the user didn't have to provide the refflat file if he/she has the annotation in gtf format

PR #124, opened 2017-05-30
Adding a tool to replace contigs in gtf files

ffinfo:


ffinfo:
Closing this PR, still need some work

ffinfo:
Tool does work like it should now, ready for merge now

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
```diff
@@ -124,8 +186,12 @@ object GtfToRefflat extends ToolCommand {
         if (gene.strand) "+" else "-",
         (transcript.transcriptionStart - 1).toString, //TODO: check if this is correct
         transcript.transcriptionEnd.toString,
-        transcript.codingStart.toString, //TODO: check if this is correct
-        transcript.codingEnd.toString, //TODO: check if this is correct
+        transcript.codingStart
+          .getOrElse(transcript.transcriptionEnd)
```
rhpvorderman, line 190: transcript.codingStart defaults to transcript.transcriptionEnd. Shouldn't that be`transcript.transcriptionStart`?

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/ReplaceContigsGtfFile.scala
```diff
@@ -0,0 +1,85 @@
...
+object ReplaceContigsGtfFile extends ToolCommand {
+  case class Args(input: File = null,
+                  output: File = null,
+                  contigs: Map[String, String] = Map(),
+                  writeAsGff: Boolean = false)
+      extends AbstractArgs
+
+  class OptParser extends AbstractOptParser {
+    opt[File]('I', "input") required () valueName "<file>" action { (x, c) =>
+      c.copy(input = x)
```
rhpvorderman, line 22: Why big letter `I` for input and small letter `o` for output? Are there any standards for this in biopet?

rhpvorderman (changes requested):

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/GtfToRefflat.scala
ffinfo: Nope, if codingstart is empty it means that there is no coding region. In that case both coding positions should be set on the end of the transcript.

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/ReplaceContigsGtfFile.scala
ffinfo: Depends who did make the tool. The standard here I copied from GATK back then but there are also tools where we do this differently.
For now this can stay like this but next meeting we can discuss a plan for this to standardize this. (will be next week)

rhpvorderman (approved):

PR #125, opened 2017-05-31
Fixing Rscripts

ffinfo:


biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/rscript/Rscript.scala
```diff
@@ -36,15 +36,15 @@ trait Rscript extends Configurable {
...
+  def cmd: Seq[String] = Seq(rscriptExecutable, Rscript.alreadyCopied(script).getAbsolutePath)
 
   /**
     * If script not exist in file system it try to copy it from the jar
     * @param dir Directory to store temp script, if None or not given File.createTempFile is called
     */
   protected def checkScript(dir: Option[File] = None): Unit = {
     if (Rscript.alreadyCopied.contains(script)) {
-      script = script.getAbsoluteFile
+      //script = script.getAbsoluteFile
```
sndrtj: Probably better to do 
```
if (!Rscript.alreadyCopied.contains(script)) { ... }
```

in stead of 

```
if (Rscript.alreadyCopied.contains(script)) {
// nothing
} else { doSomething } 

```

sndrtj (changes requested):

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/rscript/Rscript.scala
ffinfo: your right, fixed that

sndrtj (approved):

PR #127, opened 2017-05-31
add option in flexiprep to downsample input with a fraction

sndrtj:


flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/QcCommand.scala
```diff
@@ -43,10 +43,20 @@ class QcCommand(val parent: Configurable, val fastqc: Fastqc, val read: String)
...
 
+  var downSampleFraction: Option[Float] =
+    config("downsample_fraction", namespace = "flexiprep", default = None)
+
   override def defaultCoreMemory = 2.0
   override def defaultThreads = 3
 
   val seqtk = new SeqtkSeq(parent)
+  val seqtkSample: Option[SeqtkSample] = downSampleFraction match {
+    case Some(f) if 0.0 < f && f < 1.0 =>
```
ffinfo, line 55: Maybe also add something like this after this case?
``` scala
case Some(f) => 
  Logging.addError("downsample_fraction should be more then 0.0 and lower then 1.0")
```

flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/QcCommand.scala
```diff
@@ -99,7 +109,15 @@ class QcCommand(val parent: Configurable, val fastqc: Fastqc, val read: String)
   }
 
   override def beforeCmd(): Unit = {
-    seqtk.input = input
+    seqtkSample match {
+      case Some(subsample) =>
+        subsample.input = input
+        subsample.output = new File(output.getParentFile, input.getName + ".subsample.fq")
+        subsample.isIntermediate = true
```
ffinfo: This is executed inside a pipe so this is redundant

ffinfo (changes requested):

ffinfo:
An generic note, could you also add this to functional tests in flexiprep?

flexiprep/src/main/scala/nl/lumc/sasc/biopet/pipelines/flexiprep/QcCommand.scala
sndrtj: :+1:

ffinfo (approved):

PR #136, opened 2017-06-08
Adding base pipeline to auto check annotations

ffinfo:


generate-indexes/src/main/scala/nl/lumc/sasc/biopet/pipelines/generateindexes/ValidateAnnotations.scala
```diff
@@ -0,0 +1,144 @@
...
+
+  @Argument(required = true)
+  var speciesDir: File = _
+
+  override def defaults = Map("abort_on_error" -> false)
+
+  /** Init for pipeline */
+  def init(): Unit = {
+    require(speciesDir.exists(), s"speciesDir does not exist: ${speciesDir.getAbsolutePath}")
+    require(speciesDir.isDirectory, "speciesDir is not a directory")
```
rhpvorderman:
```scala
 require(speciesDir.isDirectory, "speciesDir is not a directory : ${speciesDir.getAbsolutePath}"
```
Might be useful as well

rhpvorderman (approved):

generate-indexes/src/main/scala/nl/lumc/sasc/biopet/pipelines/generateindexes/ValidateAnnotations.scala
ffinfo: agree, added this

PR #143, opened 2017-06-19
Adding intervals as a config value for users

ffinfo:


akaljuvee:
a question - does this mean that for the use case you add this, VcfStats isn't called from ShivaVariantCalling? If it would be so, then the param 'regions_of_interest' could be used for setting intervals. Or the idea is to disable running VcfStats without the intervals?

akaljuvee (approved):

ffinfo:
The option `regions_of_interest` is only to do a additional vcfstats. The one on the complete genome is still also being executed. The problem with that is with some reference vcfstats has some issue in run time because of the number of contigs. As workaround supping a bed file to limit the number of contigs this config value is needed.

I will still make an issue for vcfstats itself to fix this in the future.

PR #146, opened 2017-06-21
adding MuTect2 to Biopet

akaljuvee:


akaljuvee:
Decided still to skip writing unit tests as the advice yesterday was not to put time into this, but at least there's one test in functional testing for it.

For me the most interesting part was to decipher what the parameters of MuTect2 do. I wrote code comments for them in nl.lumc.sasc.biopet.extensions.gatk.MuTect2. These could be moved to Biopet's documentation, if this would be of use in the future, if so then the explanations could be made better. But if that's not the case, if everyone knows about it already, then let them be.

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/BiopetQScript.scala
```diff
@@ -139,7 +139,7 @@ trait BiopetQScript extends Configurable with GatkLogging { qscript: QScript =>
           if (f.getClass.isAnonymousClass) f.getClass.getSuperclass.getSimpleName
           else f.getClass.getSimpleName
         BiopetQScript.safeOutputs(f) match {
-          case Some(o) =>
+          case Some(o) if o.nonEmpty =>
```
ffinfo, line 142: Doing this is not safe. Still this is fine but there should be an extra case like this:
```scala
           case Some(_) => 
             f.jobOutputFile = new File("./stdout") // Line is here for test backup
             Logging.addError(s"on $f are no output files found")
           case _ => f.jobOutputFile = new File("./stdout") // Line is here for test backup
```

Is required for Queue to always have 1 output files for each job.

basty/src/main/scala/nl/lumc/sasc/biopet/pipelines/basty/Basty.scala
```diff
@@ -210,7 +210,7 @@ class Basty(val parent: Configurable) extends QScript with MultiSampleQScript {
                        snpsOnly: Boolean = false): FastaOutput = {
     val bastyGenerateFasta = new BastyGenerateFasta(this)
     bastyGenerateFasta.outputName = if (outputName != null) outputName else sampleName
-    bastyGenerateFasta.inputVcf = shiva.multisampleVariantCalling.get.finalFile
+    bastyGenerateFasta.inputVcf = shiva.multisampleVariantCalling.get.finalFile.get // appropriate checks have already been done when initializing Shiva
```
ffinfo: Shiva could start with only mutect2 or not? That means that this get can give a None.get error. This is not something Shiva will check.

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+  setupScatterFunction = { case scatter: GATKScatterFunction => scatter.includeUnmapped = false }
+
+  def analysis_type: String = "MuTect2"
+
+  /** Bam file for the tumor sample. */
+  @Input(fullName = "tumor_bam", required = true)
+  var tumorSampleBam: File = _
+
+  /** Bam file for the normal sample. */
+  @Input(fullName = "normal_bam", required = true)
```
ffinfo: It is possible to run mutect2 without a normal sample. This might be not what they recommencement but it's still possible

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+
+class MuTect2(val parent: Configurable) extends CommandLineGATK with ScatterGatherableFunction {
+  scatterClass = classOf[LocusScatterFunction]
+  setupScatterFunction = { case scatter: GATKScatterFunction => scatter.includeUnmapped = false }
+
+  def analysis_type: String = "MuTect2"
+
+  /** Bam file for the tumor sample. */
+  @Input(fullName = "tumor_bam", required = true)
+  var tumorSampleBam: File = _
```
ffinfo: input should but 1 list of samples. Tumor or normal can be done with the TaggedFile class. Also input files are already defined in the general extension from gatk so should not be defined at all in mutect2

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+  /** vcf file with info from cosmic db TODO desc  */
+  @Input(fullName = "cosmic", shortName = "cosmic", required = false)
+  var cosmic: Option[File] = config("cosmic")
+
+  /** Vcf file of the dbSNP database. When it's provided, then it's possible to use the param 'dbsnpNormalLod', see the
+    * description of that parameter for explanation. In addition, sIDs from this file are used to populate the ID column
+    * of the output.
+    * */
+  @Input(fullName = "dbsnp", shortName = "D", required = false)
+  var dbsnp: Option[File] = if(config("use_dbsnp", default = false)) dbsnpVcfFile else None
```
ffinfo: only use it like this: 
```scala
val dbsnp: Option[File] = config("dbsnp_vcf")
```
It's already optional

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+  @Input(fullName = "contamination_fraction_per_sample_file", shortName = "contaminationFile", required = false)
+  var contaminationFile: Option[File] = config("contamination_file")
+
+  /** Output file of the program. */
+  @Output(fullName = "out", shortName = "o", required = true)
+  @Gather(classOf[CatVariantsGatherer])
+  var outputVcf: File = _
+
+  /**
+    * The very first threshold value that is used when assessing if a particular site is a variant in the tumor sample.
```
ffinfo, line 54: We only running the tool to what is does is not directly needed inside a extention. This can also be changed in versions

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+    * classified as a somatic variant existing only in tumor. The threshold is applied on the logarithm of the ratio
+    * between the probabilities of the model that normal doesn't have a variant at the site and the probability
+    * of the model that it has a variant.
+    *
+    * Default value: 2.2
+    * */
+  @Argument(fullName = "normal_lod", required = false)
+  var normalLOD: Option[Double] = None
+
+  /** TODO: unclear how it differs from the param above 'normalLod'.
```
ffinfo, line 84: See comment above, this can be removed. If you want to know this this should be submitted to the gatk forum

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
```diff
@@ -0,0 +1,233 @@
...
+  var enableStrandArtifactFilter: Boolean = config("enable_strand_artifact_filter", default = false)
+
+  /** TODO */
+  @Argument(fullName = "useNewAFCalculator", shortName = "newQual", required = false)
+  var useNewAFCalculator: Boolean = config("use_new_af_calculator", default = false)
+
+
+  override def cmdLine = super.cmdLine +
+    required("-I:tumor", tumorSampleBam) +
+    required("-I:normal", normalSampleBam) +
```
ffinfo: the -I can be removed here, see also comments above for this

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/IoUtils.scala
```diff
@@ -85,6 +85,13 @@ object IoUtils {
     lines
   }
 
+  def writeLinesToFile(output: File, lines: List[String]): Unit = {
+    val writer = new PrintWriter(output)
+    lines.foreach(writer.println(_))
+    writer.flush()
```
ffinfo: Seems useless when closing afterwards

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/Logging.scala
```diff
@@ -38,10 +38,10 @@ object Logging {
 
   private[biopet] val errors: ListBuffer[Exception] = ListBuffer()
 
-  def addError(error: String, debug: String = null): Unit = {
+  def addError(error: String, debug: String = null, cause: Exception = null): Unit = {
```
ffinfo, line 41: Please don't use null, convert this to a option

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
```diff
@@ -287,7 +302,12 @@ class Shiva(val parent: Configurable)
 
       annotation.foreach { toucan =>
         toucan.outputDir = new File(outputDir, "annotation")
-        toucan.inputVcf = vc.finalFile
+        toucan.inputVcf = vc.finalFile.get
+        add(toucan)
+      }
+      annotationForSomaticVariants.foreach { toucan =>
+        toucan.outputDir = new File(outputDir, "annotation")
```
ffinfo, line 309: This should not be the same directory

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
```diff
@@ -152,20 +181,36 @@ trait ShivaReportTrait extends MultisampleMappingReportTrait {
...
-          "SNV Calling" -> ReportSection(
-            "/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariants.ssp"))
-      else Nil
-    val svSection =
-      if (svCallingExecuted)
-        List(
-          "SV Calling" -> ReportSection(
-            "/nl/lumc/sasc/biopet/pipelines/shiva/sampleVariantsSv.ssp"))
-      else Nil
+    var addedSections: List[(String, ReportSection)] = List()
```
ffinfo, line 184: no need to use a `var` here

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  def name = "mutect2"
+
+  override val mergeVcfResults: Boolean = false
+
+  // currently not relevant, at the moment only one somatic variant caller exists in Biopet
+  // and results from this won't be merged together with the results from other methods
+  def defaultPrio = -1
+
+  //val buildPONFromNormals: Boolean = config("build_PON_from_normals", default = false)
```
ffinfo: please remove or add some thing why this is comment out

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+
+    var renameSamples: List[String] = List()
+    var tumorSamples: List[String] = List()
+    var intermResult: File = null
+
+    if (tnPairs.size == 1) {
+      val pair = tnPairs.head
+      renameSamples = List(s"TUMOR ${pair.tumorSample}")
+      tumorSamples = List("TUMOR")
+      intermResult = new File(samplesDir, s"${pair.tumorSample}-${pair.normalSample}.$name.vcf")
```
ffinfo: As discussed before, this should be a .vcf.gz file

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+      addJobForPair(pair, intermResult)
+
+    } else {
+      var outputPerSample: List[TaggedFile] = List()
```
ffinfo: Please don't use a `var` but a `for () yield` construct

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  //val buildPONFromNormals: Boolean = config("build_PON_from_normals", default = false)
+
+  val runConEst = config("run_contest", default = "false")
+
+  def biopetScript(): Unit = {
+    val samplesDir: File = new File(outputDir, "samples")
+    if (!samplesDir.exists()) samplesDir.mkdir()
+
+    var renameSamples: List[String] = List()
```
ffinfo, line 26: Also no need for those vars here

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+
+    add(IGVToolsIndex(this, intermResult))
+    add(Tabix(this, outputFile))
+
+  }
+
+  def addJobForPair(pair: TumorNormalPair, outFile: File): Unit = {
+    val muTect2 =
+      gatk.MuTect2(this, inputBams(pair.tumorSample), inputBams(pair.normalSample), outFile)
+    // TODO add also BQSR file?
```
ffinfo: As already told you, yes this is required

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+    selectVariants.variant = intermResult
+    selectVariants.sample_name = tumorSamples
+
+    val file: File = new File(config("output_dir").asFile, ".renameSamples.txt")
+    file.deleteOnExit()
+    IoUtils.writeLinesToFile(file, renameSamples)
+
+    add(selectVariants | BcftoolsReheader(this, file) | new Bgzip(this) > outputFile)
+
+    add(IGVToolsIndex(this, intermResult))
```
ffinfo: because all outputs should be a .vcf.gz this tool should be replaced by tabix

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+      for (pair <- tnPairs) {
+        val pairLabel = s"${pair.tumorSample}-${pair.normalSample}"
+        val out: File = new File(samplesDir, s"$pairLabel.$name.vcf")
+        renameSamples :+= s"TUMOR.$pairLabel ${pair.tumorSample}"
+        tumorSamples :+= s"TUMOR.$pairLabel"
+        outputPerSample :+= TaggedFile(out, pairLabel)
+        addJobForPair(pair, out)
+      }
+
+      var sIndex = outputFile.getAbsolutePath.lastIndexOf(".vcf")
```
ffinfo: No idea what your doing here but I think this part should really be redone completly

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,101 @@
...
+
+  override val mergeVcfResults: Boolean = false
+
+  // currently not relevant, at the moment only one somatic variant caller exists in Biopet
+  // and results from this won't be merged together with the results from other methods
+  def defaultPrio = -1
+
+  //val buildPONFromNormals: Boolean = config("build_PON_from_normals", default = false)
+
+  val runConEst = config("run_contest", default = "false")
```
ffinfo: This should be: 
```scala
val runConEst: Boolean = config("run_contest", default = false)
```

ffinfo (changes requested):

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
akaljuvee: I don't understand why we should ignore their statements that tumor-only calling is not supported. This is clearly written in the manual for MuTect2 and the Best Practices page for the tool also recommends using it only with tumor-normal pairs.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
akaljuvee: but it is needed, the operator ':+' that I use, doesn't modify the existing list but creates a new one

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
akaljuvee: Are you sure about this? The name of the output file from Toucan is decided based on the input file, so the file names for output from 'annotation' and 'annotationForSomaticVariants' will differ.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
akaljuvee: For understanding what's going on, you should look a bit further downwards. The names of the tumor samples will be used by SelectVariants - so that there would be a file containing only tumor samples. The info that is gathered to the list 'renameSamples' will be used as an input for 'BcftoolsReheader', for renaming samples in the vcf file. After that the sample names in the vcf file will be the original ones, those that the user has given.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
akaljuvee: The input file for SelectVariants has to be in vcf format.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
ffinfo: Not sure but also not sure that those files names are always correct. Like the summary .out and .done files are the same name, this causing issues in the graph

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaReport.scala
ffinfo: ye but it is using the reference what is exactly the problem

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
ffinfo: nope, selectVariants accepts .vcf.gz, .vcf, .bcf and .bcf.gz

akaljuvee: nl.lumc.sasc.biopet.extensions.gatk.SelectVariant, line 30: /** Input VCF file */
and in GATK's documentation:
"--variant / -V
Input VCF file
Variants from this VCF file are used by this tool as input. The file must at least contain the standard VCF header lines, but can be empty (i.e., no variants are contained in the file).
This argument supports reference-ordered data (ROD) files in the following formats: BCF2, VCF, VCF3"

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/MuTect2.scala
akaljuvee: I have different opinion about this. To my mind only the wrapper should know what is the exact syntax for calling GATK's tool, it's not good to spread this elsewhere, that the caller of the wrapper should also know this. With nl.lumc.sasc.biopet.extensions.gatk.ContEst it's the same, why should the caller know that the tumor sample should have the tag 'eval' and that the normal sample has the tag 'genotype'. But indeed those vars could be removed and the input_file var from the superclass could be used instead, but that without requiring that the caller would have to use TaggedFile-s. Are you ok with this solution?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
akaljuvee: If I understand correct, then running first GATK's PrintReads with the bqsr-file, and then MuTect2 with the modified bam, will give the same result. I think it would be better to stick to that option. The reason is that MuTect2 expects one bqsr-file and that means that BaseRecalibrator should be run with 2 bams (found this from their forum, link below). And this would mean making changes in Biopet in how BaseRecalibrator is called, and I'd prefer not to change that. Cause what is needed is actually there already - PrintReads can be used.

http://gatkforums.broadinstitute.org/gatk/discussion/8036/mutect2-with-bqsr-flag

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/Logging.scala
akaljuvee: I don't see that using null would do any harm here. Do you want also the string to be an Option, this would mean changing this everywhere in the code, debug to Some(debug), I'd prefer it to be null as it was before.

akaljuvee:
I'm done now with the changes for this

PR #148, opened 2017-07-03
Fix fastq sync prefixes

ffinfo:


biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/FastqSync.scala
```diff
@@ -183,4 +193,21 @@ object FastqSync extends ToolCommand {
       BWriter.close()
     }
   }
+
+  def findR1R2Subfixes(fastqR1: File, fastqR2: File): (String, String) = {
```
sndrtj: Could you please add a docstring as to what this does.

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/FastqSync.scala
```diff
@@ -164,7 +172,9 @@ object FastqSync extends ToolCommand {
 
     val commandArgs: Args = parseArgs(args)
 
-    val refReader = new FastqReader(commandArgs.refFastq)
+    val subFixes = findR1R2Subfixes(commandArgs.refFastq1, commandArgs.refFastq2)
```
sndrtj: This value does not seem to be used anywhere.

sndrtj (changes requested):

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/FastqSync.scala
ffinfo: Good point, something went wrong there

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/FastqSync.scala
ffinfo: Added this

PR #150, opened 2017-07-06
First implementation of PIM in biopet

ffinfo:


biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/pim/Pim.scala
```diff
@@ -0,0 +1,103 @@
...
+    "id" -> id,
+    "description" -> description,
+    "parent_group" -> parentGroup
+  )
+}
+case class Node(id: String,
+                groupId: String,
+                inPorts: List[Port],
+                outPorts: List[Port],
+                nodeType: String)
```
bow, line 53: Aren't `nodeType` or other type arguments better as enums or case classes?

biopet-utils/pom.xml
```diff
@@ -31,6 +31,11 @@
 
     <dependencies>
         <dependency>
+            <groupId>com.typesafe.play</groupId>
```
bow, line 34: Isn't the play framework a little too big for just making HTTP calls? AFAIK there are smaller libraries (e.g. [Dispatch](https://dispatchhttp.org/Dispatch.html))

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/pipelinestatus/PipelineStatus.scala
```diff
@@ -68,15 +81,40 @@ object PipelineStatus extends ToolCommand {
...
+    val pimRunId =
+      if (cmdArgs.pimHost.isDefined) Some(cmdArgs.pimRunId.getOrElse {
+        val graphDir = depsFile.getAbsoluteFile.getParentFile
+        if (graphDir.getName == "graph") "biopet_" + graphDir.getParentFile.getName
+        else "biopet_" + depsFile.getAbsolutePath.replaceAll("/", "_")
+      })
+      else None
+
+    if (cmdArgs.pimHost.isDefined) {
+      require(pimRunId.isDefined, "Could not auto genrate Pim run ID, please supply --pimRunId")
```
bow: `s/auto genrate/auto-generate/g`

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/WriteDependencies.scala
```diff
@@ -16,10 +16,14 @@ package nl.lumc.sasc.biopet.core
 
 import java.io.{File, PrintWriter}
 
+import akka.actor.ActorSystem
```
bow, line 19: Is the entire akka framework strictly necessary here? We are only doing HTTP calls right?

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/pipelinestatus/Deps.scala
```diff
@@ -0,0 +1,60 @@
...
+import java.io.File
+
+import nl.lumc.sasc.biopet.utils.ConfigUtils
+
+/**
+  * Created by pjvanthof on 24/06/2017.
+  */
+case class Deps(jobs: Map[String, Job], files: Array[Map[String, Any]]) {
+
+  def compressOnType(main: Boolean = false): Map[String, List[String]] = {
```
bow: In here and other places, Scaladoc should be added.

bow (changes requested):

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/pim/Pim.scala
ffinfo: as far I know this is a free form

biopet-utils/pom.xml
ffinfo: yes but try reading this ;)
https://www.implicitdef.com/2015/11/19/comparing-scala-http-client-libraries.html

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/pipelinestatus/PipelineStatus.scala
ffinfo: fixed

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/WriteDependencies.scala
ffinfo: Yes, this seems like overkill indeed. But also want to see if this framework can be used later to make live reporting while running. Then such a framework is needed anyway

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/pipelinestatus/Deps.scala
ffinfo: indeed, will work on that

biopet-utils/pom.xml
bow: That blog post is slightly outdated, I think. Dispatch is still quite maintained (well, a newer version of it): https://github.com/dispatch/reboot. Also, if this is still not suitable, there should be a lot of smaller Java libraries for HTTP clients right? I think there is an Apache project made for this.

ffinfo: I think I will stay with play for now. I do agree that smaller liraries are nice but seeing the future plan I have the play/akka framework does suit better. I will not have 2 libraries inside biopet that overlap.

bow (approved):

PR #155, opened 2017-07-07
wrapper for MuTect2

akaljuvee:


biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/Logging.scala
```diff
@@ -38,10 +38,10 @@ object Logging {
 
   private[biopet] val errors: ListBuffer[Exception] = ListBuffer()
 
-  def addError(error: String, debug: String = null): Unit = {
+  def addError(error: String, debug: String = null, cause: Exception = null): Unit = {
```
ffinfo, line 41: Can this be changed to `Option[Exception]`?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/Shiva.scala
```diff
@@ -60,6 +61,10 @@ class Shiva(val parent: Configurable)
       override def configNamespace: String = "shivavariantcalling"
       override def configPath: List[String] = super.configPath ::: "multisample" :: Nil
       genders = samples.map { case (sampleName, sample) => sampleName -> sample.gender }
+      if (isSomaticVariantCallingConfigured()) {
```
ffinfo, line 64: I think it won't harm to load this always, no need to changed it when you don't have time ;)

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
```diff
@@ -108,7 +131,9 @@ class ShivaVariantcalling(val parent: Configurable)
               .callersList(this)
               .map(_.name)
               .mkString(", "))
-    if (!callers.exists(_.mergeVcfResults))
+    if (!isGermlineVariantCallingConfigured())
+      Logging.addError("For running the pipeline at least one germline variant caller has to be configured")
```
ffinfo: This excludes the possibility run it with only mutect2. It sould be possible to run germline alone, somatic alone and combined. So the only check that should be there is there is at least 1 variant caller, no matter what type.

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
```diff
@@ -226,5 +254,84 @@ object ShivaVariantcalling extends PipelineCommand {
...
+    }
+    result
+  }
+
+  private def validateTnPairTags(pairs: List[TumorNormalPair], validSampleNames:Set[String]): Unit = {
+    validateSampleNames(pairs, validSampleNames)
+
+    val samplePairs:Map[(String, String), List[TumorNormalPair]] = pairs.groupBy(pair =>
+      if(pair.tumorSample<pair.normalSample)
+        Tuple2(pair.tumorSample, pair.normalSample)
```
ffinfo: `Tuple2` is redundant here. It also does not harm of course ;)

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
```diff
@@ -226,5 +254,84 @@ object ShivaVariantcalling extends PipelineCommand {
       new RawVcf(root) ::
       new Bcftools(root) ::
       new BcftoolsSingleSample(root) ::
-      new VarscanCnsSingleSample(root) :: Nil
+      new VarscanCnsSingleSample(root) ::
+      new MuTect2(root) :: Nil
+
```
ffinfo, line 266: I think we should look at those sample tags, this seems not really efficient currently

biopet-extensions/src/main/scala/nl/lumc/sasc/biopet/extensions/gatk/CommandLineGATK.scala
```diff
@@ -418,3 +418,8 @@ trait CommandLineGATK extends BiopetJavaCommandLineFunction with Reference with
     optional("-l", logging_level, spaceSeparated = true, escape = true, format = "%s") +
     optional("-log", log_to_file, spaceSeparated = true, escape = true, format = "%s")
 }
+
+object CommandLineGATK {
+
+  def isFileWithTag(file: File, tag: String): Boolean = file.isInstanceOf[TaggedFile] && file.asInstanceOf[TaggedFile].tag == tag
```
ffinfo, line 424: I think this should be replaced with:
``` scala
def isFileWithTag(file: File, tag: String): Boolean = file match {
  case f:TaggedFile => f.tag == tag
  case _ => false
}
```

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/ShivaVariantcalling.scala
```diff
@@ -128,6 +153,9 @@ class ShivaVariantcalling(val parent: Configurable)
       caller.namePrefix = namePrefix
       caller.outputDir = new File(outputDir, caller.name)
       caller.genders = genders
+      if (caller.isInstanceOf[SomaticVariantCaller])
+        caller.asInstanceOf[SomaticVariantCaller].tnPairs = tnPairs
```
ffinfo, line 164: This works but the more scala way of doing this is:
``` scala
caller match {
  case c:SomaticVariantCaller => c.tnPairs = tnPairs
  case _ =>
}
```

ffinfo (changes requested):

ffinfo:
Taking over on branch fix-BIOPET-736-peter now

PR #157, opened 2017-07-10
Adding mutect2 pipeline

ffinfo:
Fixing BIOPET-736
Fixing BIOPET-742


biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/Logging.scala
```diff
@@ -38,10 +38,19 @@ object Logging {
...
-    errors.append(new Exception(msg))
+  def addError(error: String,
+               debug: Option[String] = None,
+               cause: Option[Exception] = None): Unit = {
+    val msg = error + debug.map("; " + _).getOrElse("")
+
+    cause match {
+      case Some(e) =>
+        logger.error(msg, e)
+        errors.append(new Exception(msg, e))
```
bow, line 49: Why `new Exception(msg, e)` instead of just `e` here?

shiva/src/main/scala/nl/lumc/sasc/biopet/pipelines/shiva/variantcallers/somatic/MuTect2.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  override val mergeVcfResults: Boolean = false
+
+  def defaultPrio: Int = -1
+
+  lazy val runConEst: Boolean = config("run_contest", default = false)
+
+  def biopetScript(): Unit = {
+
+    if (tnPairs.isEmpty) Logging.addError("No tumor-normal found in conig")
```
bow: `s/conig/config/g` here

bow (approved):

biopet-utils/src/main/scala/nl/lumc/sasc/biopet/utils/Logging.scala
ffinfo: Now you can give an own message instead of the exception you have get some a 3 party library, the other message is still in the stacktrace

PR #165, opened 2017-07-21
Adding spark support for VcfStats

ffinfo:


biopet-tools/src/test/scala/nl/lumc/sasc/biopet/tools/VcfStatsTest.scala
```diff
@@ -378,26 +248,22 @@ class VcfStatsTest extends TestNGSuite with Matchers {
 
     val blah = checkGeneral(record, List())
```
rhpvorderman: What does "blah" stand for?

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/TemplateTool.scala
```diff
@@ -17,7 +17,7 @@ package nl.lumc.sasc.biopet.core
 import java.io.{File, PrintWriter}
 
 import nl.lumc.sasc.biopet.utils.summary.Summary
-import nl.lumc.sasc.biopet.utils.{ConfigUtils, Question, ToolCommand}
+import nl.lumc.sasc.biopet.utils.{AbstractOptParser, ConfigUtils, Question, ToolCommand}
```
rhpvorderman, line 20: Lots of code duplication in the import statements. Is there an elegant way to make these more nested and hierarchical?
For example: RandomTool imports CommonToolImports. And CommonToolImports imports GeneralImports.

rhpvorderman (changes requested):

biopet-core/src/main/scala/nl/lumc/sasc/biopet/core/TemplateTool.scala
ffinfo: That's not possible but all we can do here is:
``` scala
import nl.lumc.sasc.biopet.utils._
```
This will include everything this package. I do think it's better to keep is like it is because this imports are mostly controlled by the IDE anyway. If they become to much it will replace it for the underscore automatically

biopet-tools/src/test/scala/nl/lumc/sasc/biopet/tools/VcfStatsTest.scala
ffinfo: nothing, in test code not all names makes sense ;).

I will change it to something else.

rhpvorderman (approved):

PR #181, opened 2017-08-14
Adding mustNotHaveVariant option

ffinfo:


rhpvorderman (commented):
Why is this function needed? Isn't the outputted boolean always the exact opposite of mustHaveVariant?

ffinfo:
That's what where I started with but this is not the case.
Because the forall statement this assumption don't hold.

```scala
samples
  .map(record.getGenotype)
  .forall(a => a.isHomRef || a.isNoCall || VcfUtils.isCompoundNoCall(a))
```

is not the same as
```scala
samples // removed the ! from the negative method
  .map(record.getGenotype)
  .exists(a => a.isHomRef || a.isNoCall || VcfUtils.isCompoundNoCall(a))
```

rhpvorderman (approved):

PR #193, opened 2017-08-29
Grouped regions to limit number of files opened

ffinfo:


biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/bamstats/BamStats.scala
```diff
@@ -182,41 +162,46 @@ object BamStats extends ToolCommand {
...
-    val stats = Await.result(Future.sequence(futures).map(_.fold(Stats())(_ += _)), Duration.Inf)
+    val totalSize = futures.size
+    val totalStats = Stats()
+    val contigStats: mutable.Map[String, Stats] = mutable.Map()
+
+    def wait(todo: List[Future[Map[BedRecord, Stats]]]): Unit = {
+      try {
+        logger.info(s"${totalSize - todo.size}/$totalSize tasks done")
+        val completed = todo.groupBy(_.isCompleted)
+        val bla = completed.getOrElse(true, Nil).foreach { f =>
```
rhpvorderman: Maybe something as foundContigStats instead of bla?

rhpvorderman (changes requested):

biopet-tools/src/main/scala/nl/lumc/sasc/biopet/tools/bamstats/BamStats.scala
ffinfo: Did remove it completely, bla was not used anyway. it currently is only a foreach loop

rhpvorderman (approved):

